PR #15, opened 2026-04-13
FastMCP alignment: new tools, prompts, and middleware

tony:
## Summary

FastMCP alignment for libtmux-mcp: new tool families, prompt recipes, middleware stack, bounded outputs, and correctness fixes.

## Breaking changes

- `search_panes` returns `SearchPanesResult` (was `list[PaneContentMatch]`). Matches move to `.matches`; new pagination fields. Migration: `for m in search_panes(...).matches`.
- Minimum `fastmcp>=3.2.4`.

## New tools

- **Discovery** — `list_servers`.
- **Waits** — `wait_for_text`, `wait_for_content_change`, `wait_for_channel`, `signal_channel`. Bounded, cancellable, emit `ctx.report_progress` / `ctx.warning`.
- **Buffers** — `load_buffer`, `paste_buffer`, `show_buffer`, `delete_buffer`. UUID-namespaced; leaked buffers GC'd on shutdown.
- **Hooks** (read-only) — `show_hook`, `show_hooks`.
- **Panes / windows** — `snapshot_pane`, `pipe_pane`, `display_message`, `paste_text`, `select_pane`, `swap_pane`, `select_window`, `move_window`, `enter_copy_mode`, `exit_copy_mode`.

## New prompts

Four recipes: `run_and_wait`, `diagnose_failing_pane`, `build_dev_workspace`, `interrupt_gracefully`. Expose as tools with `LIBTMUX_MCP_PROMPTS_AS_TOOLS=1`.

## Middleware

`TimingMiddleware`, `ErrorHandlingMiddleware`, `AuditMiddleware`, `SafetyMiddleware`, `ReadonlyRetryMiddleware`, `TailPreservingResponseLimitingMiddleware`.

## Bounded outputs

`capture_pane`, `snapshot_pane`, `show_buffer` take `max_lines` (default 500) with tail-preserving truncation. Pass `max_lines=None` to opt out.

## Fixes

- `search_panes` — neutralize tmux format-string injection.
- macOS `TMUX_TMPDIR` self-kill guard — resolve socket via `display-message` before env fallback.
- `build_dev_workspace` prompt — real parameter names, drop post-launch prompt waits, OS-neutral `log_command`.

## Test plan

- [x] `uv run ruff check . && uv run ruff format --check .`
- [x] `uv run mypy`
- [x] `uv run py.test --reruns 0` — 276 tests pass
- [x] `just build-docs`
[...]

PR #21, opened 2026-04-19
fix(wait_for_tools): unblock event loop via `asyncio.to_thread`

### LINKED ISSUE #18: wait_for_channel and signal_channel block the FastMCP event loop (sync subprocess.run)
## Summary

`wait_for_channel` and `signal_channel` are registered as synchronous (`def`, not `async def`) tools that run blocking `subprocess.run(..., timeout=N)` calls on FastMCP's event-loop thread. For the duration of the timeout the server cannot service stdio reads/writes, respond to the MCP client's keepalive pings, or run other tool handlers. Clients whose stdio keepalive is shorter than the `wait_for_channel` timeout can disconnect mid-wait.

0.1.0a2 explicitly addressed the same class of bug for `wait_for_text` / `wait_for_content_change` — per `CHANGES`, _"unblock FastMCP event loop via asyncio.to_thread around capture_pane"_ (commit `0a408fe`) — but left the channel-wait pair untouched.

## Evidence

`src/libtmux_mcp/tools/wait_for_tools.py:99-154` — `wait_for_channel` is a plain `def` decorated with the sync `@handle_tool_errors`. The body is a synchronous `subprocess.run` with a user-supplied timeout of up to 30 s by default and no upper bound:

```python
@handle_tool_errors
def wait_for_channel(
    channel: str,
    timeout: float = 30.0,
    socket_name: str | None = None,
) -> str:
    ...
    try:
        subprocess.run(argv, check=True, capture_output=True, timeout=timeout)
    except subprocess.TimeoutExpired as e:
        ...
```

`src/libtmux_mcp/tools/wait_for_tools.py:157-196` — `signal_channel` has the same structure (sync `def`, sync `subprocess.run` bounded by `_SIGNAL_TIMEOUT_SECONDS`). Signalling is normally fast, but the pattern is still wrong.

**Contrast — already-correct pattern:** `src/libtmux_mcp/tools/pane_tools/wait.py:100` declares `async def wait_for_text(...)` and offloads blocking work via `asyncio.to_thread(...)`. That is the pattern the channel pair should match.

## Reproduction

Environment: `libtmux-mcp==0.1.0a2` via `uvx libtmux-mcp==0.1.0a2`, connected to any MCP client with stdio keepalive. The event-loop block is client-agnostic; the observed-disconnect behavior in step 4 was seen with Claude Code but any client whose stdio keepalive is shorter than the `wait_for_channel` timeout is at risk.

1. Create an isolated tmux server so this doesn't pollute `default`: `tmux -L wfc_repro new-session -ds repro`.
2. From the MCP client, call `wait_for_channel` on a channel that will never be signalled, with a timeout safely larger than the client's stdio keepalive: `{"channel": "unsignalled", "timeout": 30, "socket_name": "wfc_repro"}`.
3. While the call is pending, call any other libtmux-mcp tool (e.g. `list_sessions`) from the same client. Observe it is queued — the server cannot respond until `wait_for_channel` returns or times out.
4. After the timeout fires, `ToolError` is raised. On Claude Code specifically, the client-side MCP harness marks the server as **disconnected** (`system-reminder: "The following deferred tools are no longer available (their MCP server disconnected)"`) even though the Python process is still alive per `ps`.
5. Tear down: `tmux -L wfc_repro kill-server`.

Observed during 0.1.0a2 smoke-testing with `timeout=2.0` — the **minimum** timeout at which the disconnect was seen. The disconnect is not guaranteed on every client or every call, but the event-loop block itself is 100% reproducible.

## Expected behavior

While `wait_for_channel` is pending, the server continues to service other tool calls, MCP pings, and cancellation signals — cooperative concurrency as the rest of the tool surface already provides.

## Suggested fix

Port the `asyncio.to_thread(...)` pattern used for `wait_for_text`:

```python
@handle_tool_errors_async
async def wait_for_channel(
    channel: str,
    timeout: float = 30.0,
    socket_name: str | None = None,
) -> str:
    server = _get_server(socket_name=socket_name)
    cname = _validate_channel_name(channel)
    argv = _tmux_argv(server, "wait-for", cname)
    try:
        await asyncio.to_thread(
            subprocess.run,
            argv, check=True, capture_output=True, timeout=timeout,
        )
    except subprocess.TimeoutExpired as e:
        ...
```

Same treatment for `signal_channel`. `handle_tool_errors_async` already exists in `_utils.py` (added in 0.1.0a2 per commit `a13898f`) — no new infrastructure required.

## Tests

Add regression coverage in `tests/test_wait_for_tools.py`:

1. Assert both functions are `asyncio.iscoroutinefunction(...)`.
2. Fire `wait_for_channel` with a 2 s unsignalled timeout on a dedicated event loop, and concurrently schedule an `asyncio.sleep(0)` coroutine via `asyncio.gather`. Assert the concurrent coroutine completes in ≪ 2 s — today it would block for the full timeout.
3. Consider mirroring the `CancelledError` propagation tests already present for `wait_for_text` in `tests/test_pane_tools.py` — the channel pair should respect the same contract.

## Related

- Commit `0a408fe` `mcp(fix[pane_tools/wait]): unblock FastMCP event loop via asyncio.to_thread around capture_pane` — the analogous fix for the pane-wait pair.
- Commit `a13898f` `mcp(refactor[_utils]): add handle_tool_errors_async for Context-using tools` — the decorator needed for the port.
- `CHANGES` 0.1.0a2 › _What's new › New tools › Waits_ lists channel tools alongside pane-wait tools; only the latter got the event-loop fix.

## Environment

- libtmux-mcp: 0.1.0a2 (PyPI + local HEAD)
- FastMCP: 3.2.4
- tmux: 3.6a
- Python: 3.13
- OS: Linux (WSL2)
- MCP client observed disconnecting: Claude Code

### LINKED ISSUE #19: is_caller annotation gives false positives across tmux sockets
## Summary

The `is_caller` boolean returned on every pane serialization (`PaneInfo.is_caller`) compares pane IDs without considering the tmux socket. A caller whose `TMUX_PANE` is `%1` on socket `default` will see `is_caller: true` for **any** pane `%1` on **any other** socket — the wrong pane, on the wrong server. The 0.1.0a2 caller-identity-scoping work fixed this for the self-kill gate but not for the informational annotation.

## Evidence

`src/libtmux_mcp/_utils.py:773` — inside `_serialize_pane`:

```python
is_caller=pane.pane_id == caller_pane_id if caller_pane_id else None,
```

Bare equality. No check against `_caller_is_on_server(server, caller)` (which lives a few hundred lines up at `_utils.py:147` and **does** do the socket comparison — realpath, then basename fallback). The 0.1.0a2 CHANGES entry reads _"Caller-identity scoping: self-protection is scoped to the caller's tmux socket rather than global"_ — accurate, but "self-protection" is specifically the kill-path gate, not this annotation.

`_serialize_pane` is used everywhere: `list_panes`, `split_window`, `create_session`, the `tmux://` resource tree in `resources/hierarchy.py`, and every other tool that returns `PaneInfo` / `PaneSnapshot`. So the false positive surfaces across most of the read-path tool surface.

## Reproduction

Environment: `libtmux-mcp==0.1.0a2`. Caller is an MCP agent running inside `TMUX_PANE=%1` on socket `default` (e.g. an agent invoked via Claude Code or Cursor from inside a tmux pane).

1. Create an isolated tmux server and split it so pane `%1` exists there too:
    ```bash
    tmux -L iscaller_repro new-session -ds repro
    tmux -L iscaller_repro split-window -t repro
    ```
2. From the MCP client, call `list_panes` with `socket_name="iscaller_repro"`.
3. **Observed:** one of the returned panes has `is_caller: true`, matching whichever pane happens to be numbered `%1` on the isolated server.
4. **Expected:** all panes on `iscaller_repro` have `is_caller: false` (or `null`) because the caller's actual pane `%1` is on `default`, not on `iscaller_repro`.
5. Tear down: `tmux -L iscaller_repro kill-server`.

In my smoke test the 0.1.0a1 stable MCP surface (before the 0.1.0a2 self-kill scoping fix) produced exactly this false positive, which I initially mis-attributed to a fix in 0.1.0a2. A code read confirms the informational annotation path was never updated.

## Suggested fix

`_serialize_pane` should consult `_caller_is_on_server(server, _get_caller_identity())` before falling back to the pane-ID equality check. Something like:

```python
def _serialize_pane(pane: Pane, server: Server | None = None) -> PaneInfo:
    caller = _get_caller_identity()
    on_server = _caller_is_on_server(server, caller) if server is not None else False
    is_caller: bool | None
    if caller is None:
        is_caller = None
    elif on_server and caller.pane_id == pane.pane_id:
        is_caller = True
    else:
        is_caller = False
    ...
```

The one wrinkle is that `_serialize_pane` today takes only a `Pane`. Either:
- Thread the `Server` through from every caller (larger refactor, many call sites), or
- Recover it via `pane.server` / `pane.window.server` (libtmux exposes this and most call sites already have a live `Pane` with a linked server).

The second option is the smaller change.

## Tests

Extend `tests/test_utils.py` and/or `tests/test_pane_tools.py` with a test that:

1. Starts two libtmux servers (`libtmux_test_A`, `libtmux_test_B`) via the existing pytest fixtures.
2. Creates a pane on server B that matches the caller's `TMUX_PANE` (e.g. `%1`).
3. Calls `_serialize_pane` (or `list_panes` end-to-end) against server B and asserts `is_caller is False`.

Mock or monkeypatch `_get_caller_identity` to return a deterministic caller if running outside tmux.

## Related

- `CHANGES` 0.1.0a2 › _Fixes_ › _macOS `TMUX_TMPDIR` self-kill guard_ — the guarded code path.
- `src/libtmux_mcp/_utils.py:147` `_caller_is_on_server` — the correct socket-scoped comparator; reuse it here.
- Affected tools (non-exhaustive): `list_panes`, `split_window`, `create_session`, `search_panes`, `snapshot_pane`, `select_pane`, `swap_pane`, `tmux://panes/<id>` resource.

## Environment

- libtmux-mcp: 0.1.0a2
- tmux: 3.6a
- Python: 3.13
- OS: Linux (WSL2)

### LINKED ISSUE #20: list_servers exposes dozens of leaked libtmux_test* tmux daemons after test runs
## Summary

Running the libtmux-mcp test suite (or any suite that uses libtmux's pytest plugin fixtures) leaves live `libtmux_test*` tmux servers and stale socket files behind after teardown. The new `list_servers` tool introduced in 0.1.0a2 makes this visible immediately — calling it returned **43 live** `libtmux_test*` servers in addition to the user's real `default` server. On the same machine, `ls /tmp/tmux-1000 | grep -c libtmux_test` returned **10,090** socket files, most of which point at long-dead servers.

The root cause lives in libtmux's pytest plugin; this issue captures the observation site for libtmux-mcp consumers so downstream users can find it.

## Evidence (from a clean session)

```
# MCP tool call (libtmux-mcp 0.1.0a2)
$ list_servers()
# returns 44 entries: "default" + 43 × "libtmux_test*"

# stale socket files on disk
$ ls /tmp/tmux-1000 | grep -c libtmux_test
10090
```

The disparity (43 live / 10 k on disk) means both leaks are happening:
- **Live leak:** a subset of test fixtures exits without calling `tmux kill-server`, so the tmux daemon keeps running indefinitely.
- **Dead-socket leak:** when tmux servers do exit, their socket files are not `unlink`ed, so `/tmp/tmux-1000` accumulates entries over time.

## Reproduction

1. Clone `tmux-python/libtmux-mcp`, check out `main`, `uv sync`.
2. Baseline count: `ls /tmp/tmux-$(id -u) | grep -c libtmux_test` (note the number, `N0`).
3. Run the full suite: `uv run pytest`.
4. Post count: `ls /tmp/tmux-$(id -u) | grep -c libtmux_test` — observe the count has **grown** past `N0` and is not cleaned up when `pytest` exits.
5. Post-live-servers: `pgrep -af "tmux.*libtmux_test"` — observe one or more live tmux daemons on `libtmux_test*` sockets.

You can also reproduce more directly with a single fixture-using test file (e.g. `uv run pytest tests/test_pane_tools.py -x -q`) and watch `ls /tmp/tmux-*/libtmux_test*` grow by a handful of files per run.

## Expected behavior

After `pytest` exits cleanly, **zero** new `libtmux_test*` processes remain and **zero** new `libtmux_test*` socket files remain under `/tmp/tmux-<uid>/`.

## Mitigation on existing machines

Reap right now:

```bash
# kill live libtmux_test* daemons
for f in /tmp/tmux-$(id -u)/libtmux_test*; do
  name=$(basename "$f")
  tmux -L "$name" kill-server 2>/dev/null
done
# remove dead socket files:
find /tmp/tmux-$(id -u) -name 'libtmux_test*' -type s -delete
```

## Related

- `CHANGES` 0.1.0a2 › _What's new › New tools › Discovery › `list-servers`_ — this tool is how the leak became visible without any bespoke tooling.
- **Root cause filed upstream**: tmux-python/libtmux#660. Fix belongs there; this issue is the observation site for libtmux-mcp consumers.

## Environment

- libtmux-mcp: 0.1.0a2
- libtmux: 0.55.x (per `pyproject.toml` floor)
- tmux: 3.6a
- Python: 3.13
- OS: Linux (WSL2)
- `/tmp/tmux-<uid>` file count at time of observation: `10090` matching `libtmux_test*`.

tony:
## Summary

- Convert `wait_for_channel` and `signal_channel` from sync `def` → `async def`.
- Wrap `subprocess.run(...)` in `await asyncio.to_thread(...)` so the FastMCP event loop is not starved for the duration of the wait.
- Swap `@handle_tool_errors` → `@handle_tool_errors_async`.

Before the fix, any MCP client whose stdio keepalive was shorter than the `wait_for_channel` timeout could disconnect mid-wait, and no other tool call could be serviced while a wait was pending. This mirrors the fix applied to `wait_for_text` / `wait_for_content_change` in 0.1.0a2 (commit 0a408fe) — the channel pair was missed.

The narrow `subprocess.TimeoutExpired` / `CalledProcessError` except blocks plus `handle_tool_errors_async`'s `Exception`-only catch mean `asyncio.CancelledError` propagates naturally — locked in as a regression test.

## Test plan

- [x] `uv run ruff check . --fix --show-fixes`
- [x] `uv run ruff format .`
- [x] `uv run mypy`
- [x] `uv run py.test --reruns 0 -vvv` (366 passed)
- [x] `just build-docs`

New regression tests in `tests/test_wait_for_tools.py`:

- `test_channel_tools_are_coroutines` — pins `async def` surface.
- `test_wait_for_channel_does_not_block_event_loop` — ticker-based discriminator: fires a 10 ms-tick coroutine in parallel with a 500 ms unsignalled wait; asserts ≥ 20 ticks (≈ 0 ticks pre-fix).
- `test_wait_for_channel_propagates_cancellation` — cancels an in-flight wait with `task.cancel()`, asserts `CancelledError` bubbles (not wrapped as `ToolError`).

Existing sync-call tests converted to `asyncio.run(...)` since the function signatures changed. One caller in `tests/test_pane_tools.py` similarly updated.

## Companion PRs

Part of a batch of post-0.1.0a2 smoke-test fixes. Each PR is independent and can land in any order:

- #21 — event-loop block in channel waits (this PR, closes #18)
- #22 — `is_caller` socket-blind false positive (closes #19)
- #23 — pytest leaks `libtmux_test*` sockets (closes #20)

Closes #18

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -26,6 +26,14 @@ _Notes on upcoming releases will be added here_
   `_caller_is_on_server` (the same socket-scoped comparator used by
   the self-kill guard) via the new `_compute_is_caller` helper
   (#22, fixes #19).
+- {tooliconl}`wait-for-channel` and {tooliconl}`signal-channel` no
+  longer block the FastMCP event loop. Both were sync `def` handlers
+  running `subprocess.run(timeout=N)` on the main loop — for the
+  duration of the wait the server could not service other tool calls,
+  MCP pings, or client cancellations. Ported to `async def` +
+  {func}`asyncio.to_thread`, matching the pattern already used by
+  {tooliconl}`wait-for-text` / {tooliconl}`wait-for-content-change`
+  (#21, fixes #18).
 
 ## libtmux-mcp 0.1.0a2 (2026-04-19)
 
--- a/src/libtmux_mcp/tools/wait_for_tools.py
+++ b/src/libtmux_mcp/tools/wait_for_tools.py
@@ -24,6 +24,7 @@ This ensures the signal fires on both success and failure paths.
 
 from __future__ import annotations
 
+import asyncio
 import re
 import subprocess
 import typing as t
@@ -35,7 +36,7 @@ from libtmux_mcp._utils import (
     TAG_MUTATING,
     _get_server,
     _tmux_argv,
-    handle_tool_errors,
+    handle_tool_errors_async,
 )
 
 if t.TYPE_CHECKING:
@@ -96,8 +97,8 @@ def _validate_channel_name(name: str) -> str:
     return name
 
 
-@handle_tool_errors
-def wait_for_channel(
+@handle_tool_errors_async
+async def wait_for_channel(
     channel: str,
     timeout: float = 30.0,
     socket_name: str | None = None,
@@ -143,7 +144,15 @@ def wait_for_channel(
     cname = _validate_channel_name(channel)
     argv = _tmux_argv(server, "wait-for", cname)
     try:
-        subprocess.run(argv, check=True, capture_output=True, timeout=timeout)
+        # FastMCP direct-awaits async tools on its event loop. ``tmux
+        # wait-for`` blocks for the full timeout (up to 30 s by default)
+        # so the synchronous ``subprocess.run`` must run off the loop —
+        # otherwise no other tool call, MCP ping, or cancellation can
+        # be serviced for the duration of the wait. Mirror the pattern
+        # used by :func:`~libtmux_mcp.tools.pane_tools.wait.wait_for_text`.
+        await asyncio.to_thread(
+            subprocess.run, argv, check=True, capture_output=True, timeout=timeout
+        )
     except subprocess.TimeoutExpired as e:
         msg = f"wait-for timeout: channel {cname!r} was not signalled within {timeout}s"
         raise ToolError(msg) from e
@@ -154,8 +163,8 @@ def wait_for_channel(
     return f"Channel {cname!r} was signalled"
 
 
-@handle_tool_errors
-def signal_channel(
+@handle_tool_errors_async
+async def signal_channel(
     channel: str,
     socket_name: str | None = None,
 ) -> str:
@@ -180,8 +189,12 @@ def signal_channel(
     cname = _validate_channel_name(channel)
     argv = _tmux_argv(server, "wait-for", "-S", cname)
     try:
-        subprocess.run(
-            argv, check=True, capture_output=True, timeout=_SIGNAL_TIMEOUT_SECONDS
+        await asyncio.to_thread(
+            subprocess.run,
+            argv,
+            check=True,
+            capture_output=True,
+            timeout=_SIGNAL_TIMEOUT_SECONDS,
         )
     except subprocess.TimeoutExpired as e:
         msg = (
--- a/tests/test_pane_tools.py
+++ b/tests/test_pane_tools.py
@@ -742,6 +742,7 @@ def test_search_panes_per_pane_matched_lines_cap(
     ``capture_pane`` (command-line plus output-line for each), well
     past the truncation threshold of three.
     """
+    import asyncio
     import uuid
 
     from libtmux_mcp.tools.wait_for_tools import wait_for_channel
@@ -753,7 +754,11 @@ def test_search_panes_per_pane_matched_lines_cap(
         f"tmux wait-for -S {channel}"
     )
     mcp_pane.send_keys(payload, enter=True)
-    wait_for_channel(channel=channel, timeout=5.0, socket_name=mcp_server.socket_name)
+    asyncio.run(
+        wait_for_channel(
+            channel=channel, timeout=5.0, socket_name=mcp_server.socket_name
+        )
+    )
 
     result = search_panes(
         pattern=marker,
--- a/tests/test_wait_for_tools.py
+++ b/tests/test_wait_for_tools.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import threading
 import time
 import typing as t
@@ -38,6 +39,20 @@ def test_validate_channel_name_rejects_invalid(name: str) -> None:
         _validate_channel_name(name)
 
 
+def test_channel_tools_are_coroutines() -> None:
+    """Both tools must be ``async def`` so FastMCP awaits them.
+
+    Regression guard for tmux-python/libtmux-mcp#18: sync ``def`` tools
+    were direct-called on FastMCP's event loop and the internal
+    ``subprocess.run`` blocked stdio for the full timeout window. The
+    fix converts both to ``async def`` + ``asyncio.to_thread``; this
+    assertion pins the async surface so a silent revert doesn't sneak
+    through.
+    """
+    assert asyncio.iscoroutinefunction(wait_for_channel)
+    assert asyncio.iscoroutinefunction(signal_channel)
+
+
 @pytest.mark.usefixtures("mcp_session")
 def test_signal_channel_no_waiter_is_noop(mcp_server: Server) -> None:
     """``tmux wait-for -S`` on an unwaited channel returns successfully.
@@ -47,9 +62,11 @@ def test_signal_channel_no_waiter_is_noop(mcp_server: Server) -> None:
     unstarted Server instance, so ``mcp_session`` is what actually
     boots the tmux process.
     """
-    result = signal_channel(
-        channel="wf_test_noop",
-        socket_name=mcp_server.socket_name,
+    result = asyncio.run(
+        signal_channel(
+            channel="wf_test_noop",
+            socket_name=mcp_server.socket_name,
+        )
     )
     assert "signalled" in result
 
@@ -61,15 +78,17 @@ def test_wait_for_channel_returns_when_signalled(mcp_server: Server) -> None:
 
     def _signal_after_delay() -> None:
         time.sleep(0.3)
-        signal_channel(channel=channel, socket_name=mcp_server.socket_name)
+        asyncio.run(signal_channel(channel=channel, socket_name=mcp_server.socket_name))
 
     thread = threading.Thread(target=_signal_after_delay)
     thread.start()
     try:
-        result = wait_for_channel(
-            channel=channel,
-            timeout=5.0,
-            socket_name=mcp_server.socket_name,
+        result = asyncio.run(
+            wait_for_channel(
+                channel=channel,
+                timeout=5.0,
+                socket_name=mcp_server.socket_name,
+            )
         )
         assert "signalled" in result
     finally:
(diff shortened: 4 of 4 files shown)


tony:
### Code review

No issues found. Checked for bugs and AGENTS.md compliance (repo has AGENTS.md, no CLAUDE.md).

Scope:
- AGENTS.md namespace-import rule and testing conventions.
- Shallow bug scan on the diff (decorator swap, `asyncio.to_thread` hand-off, exception propagation).
- Git blame / history — verified the change matches the pattern established by commit `0a408fe` (the analogous `wait_for_text` async fix) and that `handle_tool_errors_async` at [`src/libtmux_mcp/_utils.py`](https://github.com/tmux-python/libtmux-mcp/blob/5b919b6ca09f9a50dec998287939d0a814b2801e/src/libtmux_mcp/_utils.py#L847-L870) correctly preserves the `CancelledError`-propagation contract locked in by the new test.
- Prior PR review comments on `wait_for_tools.py` / `pane_tools/wait.py` (PRs #11, #15) — no applicable guidance.
- Code comments in modified files — the `Safety` docstring's timeout invariant at [`src/libtmux_mcp/tools/wait_for_tools.py`](https://github.com/tmux-python/libtmux-mcp/blob/5b919b6ca09f9a50dec998287939d0a814b2801e/src/libtmux_mcp/tools/wait_for_tools.py#L11-L22) is preserved.

PR #25, opened 2026-04-19
fix(server): guide clients to emit JSON-quoted string arguments

### LINKED ISSUE #17: MCP tool calls fail: invalid JSON when passing session_name / session_id (Cursor agent)
## Summary

When driving **libtmux-mcp** from **Cursor’s agent**, some tool invocations fail **before** the Python server runs: the MCP layer rejects the request because the JSON payload for tool arguments is **malformed**.

This is **not** the same as the existing workaround for dict-shaped parameters sent as JSON **strings** (`_coerce_dict_arg` / Composer dict-stringification). Here, **string values that must be JSON strings are emitted without quotes**, producing invalid JSON.

## Symptoms

Errors observed on the client/MCP bridge when calling tools that accept optional session targeting, for example:

- `Unexpected token '$', "{"session_id": $10}" is not valid JSON`
- `Unexpected token 'c', "{"session_name": cv}" is not valid JSON`

Valid JSON would require:

- `"session_id": "$10"`
- `"session_name": "cv"`

## Impact

- Tools such as **`list_windows`** (with `session_name` / `session_id`) and **`select_window`** fail entirely; the agent falls back to shell `tmux` commands.
- Any tool parameter that is a **string** could theoretically hit the same class of bug if the serializer mishandles it.

## Steps to reproduce (agent)

1. Use Cursor with the libtmux MCP server configured (stdio).
2. From the agent, call **`list_windows`** with `session_name` set to a bare session name (e.g. `cv`), or **`list_windows`** / related tools with `session_id` (e.g. `$10`).
3. Observe MCP parse/validation errors instead of a normal tool result.

(Exact UI steps depend on Cursor version; the failure is visible in the agent tool-call error output.)

## Expected behavior

Tool arguments are serialized as **valid JSON**. String parameters are JSON-encoded (quoted, with proper escaping).

## Actual behavior

The serialized arguments are **not valid JSON** (unquoted `$10`, unquoted `cv`), so the request fails at parse time.

## Nature of the bug

- **Likely location:** MCP client / Cursor’s tool-argument serialization for this model or code path—not libtmux-mcp’s Python handlers (they never run).
- **Relation to existing work:** The project already documents **dict** parameters arriving as JSON **strings** (Composer 1/1.5) in `_coerce_dict_arg` and `create_session.environment`. This issue is **orthogonal**: it is **invalid JSON at the wire level**, not a semantic coercion problem.

## Suggestions

1. **Upstream:** If reproducible only in Cursor, report or cross-link to Cursor with a minimal failing tool call payload.
2. **Docs:** Optionally note in MCP/client docs that agents should prefer **globally unique `pane_id`** targeting where possible to reduce reliance on session filters.
3. **Investigation:** Confirm whether FastMCP / MCP server can surface a clearer error when the client sends malformed JSON (may be outside server control).

## Environment (observed)

- OS: Linux (WSL2)
- Cursor agent + libtmux-mcp MCP server (stdio)
- tmux sessions present (e.g. session name `cv`, id `$10`)

---

If maintainers prefer this tracked as **documentation-only** or **upstream-meta**, feel free to relabel; the goal is to capture the failure mode and distinguish it from `_coerce_dict_arg`-style workarounds.

tony:
## Summary

Adds an explicit `JSON ARGUMENT FORMAT` paragraph to the server's MCP `instructions` payload, telling the client's model to emit JSON-quoted string values for tool arguments — a remediation for the Cursor Auto Mode bug reported in #17 where string values like `"$10"` (`session_id`) and `"cv"` (`session_name`) were sent **unquoted**, producing invalid JSON that the MCP stdio transport rejected before any server handler ran.

## Why prompt-engineering rather than a wire-level shim

Two honest options on the table:

1. **Prompt-engineering (this PR)** — add an instruction to the server's `instructions=` channel. Clean, lives in the natural MCP API surface, no monkey-patching. Downside: model obedience is probabilistic.
2. **Wire-level shim** — monkey-patch `mcp.types.JSONRPCMessage.model_validate_json` to regex-repair bare tokens on parse failure. Deterministic, but version-fragile and invasive.

Shipping (1) first. If real-world Cursor usage still shows the bug after this lands (i.e. Cursor's Auto Mode model path doesn't honor server MCP instructions), the shim is ready as a follow-up. Scoping drafted notes for that are retained locally.

## What changed

- `src/libtmux_mcp/server.py` — extends `_BASE_INSTRUCTIONS` with a JSON-ARGUMENT-FORMAT paragraph between the preamble and the metadata-vs-content paragraph. Tone matches the existing all-caps section headers (`IMPORTANT — metadata vs content`, `READ TOOLS TO PREFER`, `WAIT, DON'T POLL`).
- `docs/clients.md` — new **Known client issues** section documenting the Cursor Auto Mode symptom, the server-side mitigation, and `pane_id` as belt-and-suspenders. Uses the existing MyST `:::{warning}` style (matches `docs/index.md` admonition).
- `CHANGES` — unreleased `Fixes` entry.

## Test plan

- [x] `uv run ruff check . --fix --show-fixes` — clean
- [x] `uv run ruff format .` — clean
- [x] `uv run mypy` — Success: no issues found in 50 source files
- [x] `uv run py.test --reruns 0 -vvv` — 366 passed
- [x] `just build-docs` — build succeeded, no warnings

**End-to-end confirmation** that the new instruction actually reaches clients: drove the MCP `initialize` handshake against this branch's server via a subprocess stdin dance — the returned `instructions` payload is 2362 chars and contains the full JSON-ARGUMENT paragraph verbatim.

**Cursor IDE verification (user-side)**: `~/.cursor/mcp.json` was swapped locally to point `libtmux` at this branch via `uv --directory /home/d/work/python/libtmux-mcp run libtmux-mcp` (the baseline pinned PyPI install is preserved as `libtmux-pinned` for before/after comparison). Interactive Cursor IDE testing is still the definitive check; `cursor-agent -p` non-interactive mode does not expose MCP tools to the agent (established earlier in this session's 4-CLI smoke test).

Closes #17


tony:
Closing unmerged — server-side mitigation for #17 isn't achievable.

**What was tried**

1. Prompt instruction in the MCP `instructions` payload telling models to emit JSON-quoted strings. Shipped as commit 9b61354. Empirically insufficient: live Cursor IDE testing with Composer showed the model acknowledging issue #17 in its narration *while* serializing the next tool call with unquoted arguments.
2. Transport-layer shim (`LibtmuxMcpServer` subclass wrapping `stdio_server(stdin=...)` with regex quote-repair). Shipped as commit cef780c. Passed the server-side probe (a synthesized malformed frame piped to the server's stdin was repaired and executed), but failed the real-world test: `cursor-agent -p` still surfaced `Unexpected token` errors because Composer's malformed output is caught by Cursor's client-side `JSON.parse` of the model's tool-call text — the bad payload never reaches our stdio transport.

**Root cause**

The bug lives in Cursor's Composer models (1, 1.5, and 2 all affected). They emit tool-call arguments as text containing unquoted string values (`{"session_name": c}`), and Cursor's client code rejects the output at `JSON.parse` before any MCP transport is involved. No server-side lever — prompt-level, transport-level, schema-level — can fix a payload that never crosses the client/server boundary.

**What users should do**

- Switch Cursor's model picker to Claude or GPT (the bug is specific to Composer).
- File upstream at https://forum.cursor.com requesting a fix to Composer's tool-call JSON serialization.

Hard-resetting the branch to `origin/main` and closing without merge.

PR #27, opened 2026-04-20
Add `respawn_pane` + `get_{session,window}_info`, refactor docs extension imports

tony:
## Summary

- **Add `respawn_pane`**: restart a pane's process in place (tmux `respawn-pane -k`), preserving `pane_id` and layout — the right primitive for agent shell-recovery flows that would otherwise `kill_pane` + `split_window` and invalidate held references.
- **Add `get_session_info` / `get_window_info`**: single-object metadata reads that complete the core-hierarchy symmetry alongside existing `get_pane_info` / `get_server_info`. Callers with an ID no longer need `list_*` + filter.
- **Sharpen LLM-facing discoverability** on `display_message` (retitled "Evaluate tmux Format String", docstring now leads with what it actually does — read-only format expansion, not a UI message) and `pipe_pane` (docstring leads with the concrete `/tmp/pane.log` logging use case).
- **Expand `_BASE_INSTRUCTIONS`** with three disambiguators: hooks are read-only by design, the `load/paste/delete_buffer` lifecycle (and why there is no `list_buffers`), and the `is_caller=true` workflow for "which pane am I in?" — heads off exploratory calls for tools that don't exist.
- **Refactor the bundled Sphinx extension**: import path changes from `widgets` to `docs._ext.widgets`, removing a mypy `mypy_path` alias hack that was needed only to paper over the dual-name resolution.
- **Add `scripts/mcp_swap.py`**: dev tool that detects which agent CLIs (Claude, Codex, Cursor, Gemini) are installed and swaps their MCP config to run the local checkout editable-style (or reverts to the pre-swap state from timestamped backups). Motivated by "does this branch work on all four CLIs?" — previously a four-config hand-edit dance; now `just mcp-use-local` / `just mcp-revert`.

### New tools (`src/libtmux_mcp/tools/`)

- **`pane_tools/lifecycle.py`**: `respawn_pane(pane_id, session_name, session_id, window_id, kill=True, shell_command, start_directory, socket_name) -> PaneInfo`. `kill=True` threads `-k` (matches the recovery-flow intent); optional `shell_command` maps to the positional arg, `start_directory` maps to `-c`. Calls `pane.refresh()` so the returned `PaneInfo` carries the fresh `pane_pid` and `pane_current_command`. Registered with `ANNOTATIONS_MUTATING` / `TAG_MUTATING`.
- **`session_tools.py`**: `get_session_info(session_id, session_name, socket_name) -> SessionInfo`. Reuses `_resolve_session` / `_serialize_session` — no new helpers. Registered with `ANNOTATIONS_RO` / `TAG_READONLY`.
- **`window_tools.py`**: `get_window_info(window_id, window_index, session_name, session_id, socket_name) -> WindowInfo`. Reuses `_resolve_window` / `_serialize_window`. Inline comment memorializes the boundary: "the symmetry argument is bounded to the four-level core hierarchy — NOT a license to add `get_buffer_info` / `get_hook_info` / `get_option_info`".

### Instruction / discoverability tweaks

- **`src/libtmux_mcp/server.py`** (`_BASE_INSTRUCTIONS` + `_build_instructions`): add `HOOKS ARE READ-ONLY` block, `BUFFERS` block, and — inside the `if tmux_pane:` branch only — a sentence pointing agents at `list_panes` + `is_caller=true` filter instead of a nonexistent `whoami` tool. The workflow sentence is deliberately scoped to the in-tmux branch because it references "your pane is identified above", which is only true when the agent-context line has been emitted.
- **`pane_tools/meta.py`** (`display_message`) + **`pane_tools/__init__.py`**: retitled to `"Evaluate tmux Format String"`; docstring now leads with "Evaluate a tmux format string against a target and return the expanded value" so the name-vs-behavior mismatch (`display_message` is the tmux verb, but `-p` expands rather than displays) stops biting LLM clients.
- **`pane_tools/pipe.py`** (`pipe_pane`): summary reframed to "Log a pane's live output to a file" with an explicit `/tmp/pane.log` hint.

### Docs extension refactor

| Before | After |
|---|---|
| `docs/_ext/` on `sys.path`, imported as `widgets` | `docs/` on `sys.path`, imported as `docs._ext.widgets` |
| `mypy_path = ["docs/_ext"]` + `exclude = ["^docs/"]` juggling to avoid "source file found twice" | `exclude = ["^docs/"]` alone — `mypy_path` alias removed |
| `docs/_ext/__init__.py` missing (implicit namespace) | explicit package marker added |

- **`docs/conf.py`**: `sys.path.insert(project_root)` replaces `sys.path.insert(_EXT_DIR)`; extension listed as `docs._ext.widgets`.
- **`tests/docs/conftest.py`** + **`tests/docs/test_widgets.py`**: imports rewritten to `from docs._ext.widgets import ...` and the generated per-test `conf.py` follows suit.
- **`docs/_widgets/mcp-install/widget.html`**: comment reference updated to `docs._ext.widgets._base.make_highlight_filter`.

### Docs pages

- `docs/tools/pane/respawn-pane.md`, `docs/tools/session/get-session-info.md`, `docs/tools/window/get-window-info.md` — each with a "Use when / Avoid when / Side effects" block, a primary example, and a resolve-by-alternate-key example where relevant.
- Session, Window, and Pane index pages gain grid-item-cards and toctree entries for the new tools.
- **`README.md`**: tool catalog rows updated for Session, Window, and Pane to list the three new tools.

### Dev tooling: `scripts/mcp_swap.py`

A single-file PEP 723 script that centralizes the "point every installed CLI at my local checkout" flow. Four subcommands: `detect` (which CLIs exist), `status` (what each resolves libtmux to today), `use-local` (rewrite each config to `uv --directory <repo> run libtmux-mcp`), `revert` (restore from backups).

| CLI    | Config path                    | Format | Server key                                         |
|--------|---------------------------------|--------|----------------------------------------------------|
| Claude | `~/.claude.json`                | JSON   | `projects."<abs-repo>".mcpServers.libtmux`         |
| Codex  | `~/.codex/config.toml`          | TOML   | `[mcp_servers.libtmux]`                            |
| Cursor | `~/.cursor/mcp.json`            | JSON   | `mcpServers.libtmux`                               |
| Gemini | `~/.gemini/settings.json`       | JSON   | `mcpServers.libtmux`                               |

- **Server name + entry** auto-derived from `pyproject.toml`: `project.name` with trailing `-mcp` stripped (`libtmux-mcp` → `libtmux`), first `[project.scripts]` key for the entry command. Flags override (`--server`, `--entry`). Generalizes cleanly to any uv-managed MCP repo.
- **Format preservation**: `tomlkit` for Codex's TOML so top-level comments, sibling tables, and key order survive the round-trip; stdlib `json` for the three JSON configs (none of them rely on comments).
- **Claude's per-project keying** is handled explicitly — only the current repo's `projects."<abs-repo>"` block is rewritten; other projects' entries are untouched. Claude's full entry shape (`type`, `env`) is preserved even when empty.
- **Safety**: timestamped `.bak.mcp-swap-<ts>` backups (never overwrites existing backups); atomic tempfile + `pathlib.Path.replace`; post-write re-parse validation with rollback on parse failure; `--dry-run` prints unified diffs and writes nothing; state file at `~/.local/state/libtmux-mcp/mcp_swap.json` tracking per-CLI backup paths so `revert` handles the "added" case (Codex had no entry → `revert` removes the block, doesn't restore a phantom previous version).
- **Four `[group: 'mcp']` just recipes**: `mcp-detect`, `mcp-status`, `mcp-use-local`, `mcp-revert`.
- **`scripts/README.md`** covers usage + extension (add an entry to `CLIS` and extend the three per-CLI branches in `get_server` / `set_server` / `delete_server`).
- **`tomlkit>=0.13`** added to the dev dependency group so pytest can import the PEP 723 script.

## Design decisions

- **`respawn_pane` defaults `kill=True`**: the tool exists for recovery flows where the process is already wedged. Defaulting to `kill=False` would surface "pane is not dead" errors as the common case and push every caller to set the flag manually.
- **`-e` env not exposed on `respawn_pane`**: `set_environment` already covers that path; keeping the surface small avoids two ways to do the same thing.
- **`respawn_window` deferred**: audit §6.1 — ship `respawn_pane` first, add the window variant only if usage demands it.
- **`get_*_info` bounded to the four-level hierarchy**: server / session / window / pane. Buffers, hooks, and options have existing `show_*` / `load_*` reads; adding `get_buffer_info` etc. would bloat the surface without adding capability. Inline comment on `get_window_info` documents this so future contributors don't relitigate.
- **`is_caller` workflow sentence lives inside `_build_instructions`, not `_BASE_INSTRUCTIONS`**: the sentence references "your pane is identified above", which is only true when `TMUX_PANE` is set. Putting it in the base string would be a lie outside tmux.
- **Docs extension as a real package**: the `mypy_path` alias was a hack — mypy saw the widget modules under two names (`widgets` and `docs._ext.widgets`) and needed an `exclude` to avoid "source file found twice" during full-tree traversal. Making `docs/_ext/__init__.py` explicit and importing consistently as `docs._ext.widgets` collapses the two paths into one.
- **`mcp_swap.py` lives in `scripts/`, not as an installed entry point**: it's developer infrastructure, not product. Keeping it a PEP 723 single file means `uv run scripts/mcp_swap.py` works with zero setup, and copying the file to another MCP repo is a one-command adoption path.
- **`mcp_swap.py` replaces rather than sidecars the libtmux entry**: one `libtmux` key per CLI, not `libtmux-dev` + `libtmux`. This avoids two entries both claiming the same tool namespace and matches the user's manual "swap" flow the `.bak.libtmux-dev` backup files hinted at.

## Verification

Verify the new tools are registered:

```
rg -n "respawn_pane|get_session_info|get_window_info" src/libtmux_mcp/tools/
```

Verify the widget import refactor left no `widgets` top-level imports:

```
rg -n "^from widgets|^import widgets" src tests docs
```

Verify `mypy_path` alias is gone:

```
rg -n "mypy_path" pyproject.toml
```

Verify instruction blocks are present:

```
rg -n "HOOKS ARE READ-ONLY|BUFFERS:|whoami tool" src/libtmux_mcp/server.py
```

End-to-end test the swap script (reversible round-trip):

```
just mcp-detect
just mcp-status
just mcp-use-local --dry-run
just mcp-use-local
just mcp-revert
```

## Test plan

- [x] `test_respawn_pane_preserves_pane_id_and_refreshes_pid` — pane_id survives, pane_pid changes
- [x] `test_respawn_pane_replaces_shell_command` — `shell_command` override takes effect
- [x] `test_get_session_info` + `test_get_session_info_by_name` — resolves by id and by name
- [x] `test_get_window_info` + `test_get_window_info_by_index` — resolves by id and by index+session
- [x] `test_base_instructions_document_hook_boundary` — `HOOKS ARE READ-ONLY` + `show_hooks` + "tmux config file" present
- [x] `test_base_instructions_document_buffer_lifecycle` — load/paste/delete/BufferRef/list_buffers/clipboard history all present
- [x] `test_build_instructions_documents_is_caller_workflow_inside_tmux` — workflow sentence present with `TMUX_PANE` set, absent without
- [x] Widget tests import via `docs._ext.widgets` and the generated test `conf.py` picks them up
- [x] `tests/test_mcp_swap.py` — 12 cases: JSON round-trip (cursor/gemini), Claude per-project keying, Codex comment preservation + add-when-missing + revert-removes, unrelated-server preservation, `--dry-run` writes nothing, second-swap-is-noop, state-file cleared on full revert, `McpServerSpec` shape helpers
- [x] End-to-end swap cycle against real configs: `mcp-use-local` → backups written → all four CLIs report local → `mcp-revert` → configs byte-identical to pre-swap backups (verified via `cmp`) → `mcp-use-local` again → libtmux entries byte-identical to the prior local state (Claude's whole file hash drifts because Claude Code autosaves unrelated fields; the libtmux entry itself is stable)
- [x] `uv run mypy .` — succeeds without the `mypy_path` alias (strict, 53 files)
- [x] `uv run ruff check .` — clean
- [x] `just test` / `uv run pytest` — full suite green (387 tests)


tony:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #28, opened 2026-04-25
ci(docs[trigger]): publish from seo-packages branch

tony:
## Summary

- Add `seo-packages` to the docs workflow's `on.push.branches` list so the publish pipeline (uv sync → just html → S3 sync → CloudFront invalidate → Cloudflare purge) can be exercised end-to-end on this branch without merging to main first.
- Refresh `uv.lock`'s `exclude-newer` timestamp (artifact of adding then dropping a temporary `[tool.uv.sources]` local-path override for the in-progress gp-sphinx SEO migration; net pyproject change vs main is zero).
- Net intent: validate the upstream gp-sphinx swap (sphinxext-opengraph → gp-opengraph, plus gp-sitemap) against libtmux-mcp's live docs pipeline before tagging a gp-sphinx release.

## Test plan

- [ ] Push to `seo-packages` triggers `.github/workflows/docs.yml`
- [ ] `dorny/paths-filter` correctly gates publish to docs-relevant changes
- [ ] `uv sync` resolves cleanly from PyPI (no local-path leakage)
- [ ] `just html` builds without warnings
- [ ] S3 sync, CloudFront invalidation, and Cloudflare cache purge all succeed
- [ ] Published site reflects the build (or workflow no-ops when no docs changed)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,9 +56,9 @@ libtmux-mcp = "libtmux_mcp:main"
 [dependency-groups]
 dev = [
   # Docs
-  "gp-sphinx==0.0.1a9",
-  "sphinx-autodoc-api-style==0.0.1a9",
-  "sphinx-autodoc-fastmcp==0.0.1a9",
+  "gp-sphinx==0.0.1a10",
+  "sphinx-autodoc-api-style==0.0.1a10",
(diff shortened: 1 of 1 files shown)


tony:
Addressed via other commit

PR #35, opened 2026-05-04
scripts(mcp_swap): add --scope {user,project} for Claude two-layer config

tony:
## Summary

- **Add** `--scope {user,project}` to `scripts/mcp_swap.py`'s `use-local` and `revert` subcommands. `--scope user` rewrites Claude's top-level `mcpServers` fallback (every project without an override picks it up); `--scope project` (the default — preserves pre-flag behaviour) writes the per-project `projects[<abs>].mcpServers` entry.
- **Migrate** state-file schema to v2: keys go from bare `cli` strings to `f"{cli}:{scope}"` so a single Claude install can hold both scopes simultaneously with independent backups. v1 entries migrate transparently on load (`claude` → `("claude","project")`; others → `("<cli>","user")`).
- **Fix** a real backup-collision bug surfaced by the coexist test — two Claude swaps inside the same second produced identical backup filenames, so the second swap's backup overwrote the first's. Backup suffix now embeds the scope for Claude (non-Claude filenames stay byte-compatible with v1).
- **Document** the new flag in `scripts/README.md` with a "Scope" subsection and the layered-config explanation that motivates it.

## Motivation

`mcp_swap.py` was conservative by design — it only wrote under the current repo's per-project key in `~/.claude.json`, leaving every other project's MCP entry untouched. Side-effect: when QA-ing a libtmux-mcp branch from any non-libtmux-mcp directory (e.g. `~/work/cv`), Claude kept resolving the user-level fallback to the PyPI release. The flag adds an explicit "yes, swap the global default" path without losing the safe-by-default per-project behaviour.

### Script (`scripts/mcp_swap.py`)

- **`Scope` literal + helpers**: New `Scope = Literal["user","project"]`, plus `_normalize_scope` (coerces non-Claude scope to `"user"`), `_state_key`, and `_parse_state_key` (the v1→v2 migration shim). All centralised so the rest of the file just threads `scope` through.
- **`get_server` / `set_server` / `delete_server`**: New keyword-only `scope` parameter (default `"project"`). Claude branches gate on it; non-Claude branches ignore it. `_claude_project_node` is unchanged — only called when `scope == "project"`.
- **`load_state` / `save_state` / `clear_state`**: Switched from `dict[CLIName, SwapEntry]` to `dict[tuple[CLIName, Scope], SwapEntry]`. Migration runs in `load_state` via `_parse_state_key`; unknown keys are dropped silently.
- **`cmd_status`**: For Claude, prints separate `[claude:user]` / `[claude:project]` lines when both have entries (or one if only one exists). Non-Claude CLIs continue to print one `[<cli>]` line.
- **`cmd_use_local` / `cmd_revert`**: Read `args.scope`, normalise per CLI, key state by `(cli, scope)`. `revert` without `--scope` rolls back every recorded scope for the targeted CLIs; with `--scope`, only that one.
- **`build_parser`**: New `--scope` argument on `use-local` and `revert` with `choices=ALL_SCOPES` and `default=None`.
- **Backup suffix**: For Claude, embeds the scope to disambiguate within-second collisions. Non-Claude suffix unchanged.

### Tests (`tests/test_mcp_swap.py`)

| Test | Validates |
|---|---|
| `test_claude_user_scope_writes_top_level_mcpServers` | `--scope user` writes `mcpServers` and creates no `projects[<abs>]` node |
| `test_claude_user_scope_round_trip_restores_byte_identical` | swap → revert at user-scope yields byte-identical bytes |
| `test_claude_user_and_project_swaps_coexist_independently` | Both scopes can swap with separate backups; reverting one leaves the other intact |
| `test_legacy_v1_state_migrates_to_v2_keys` | Bare-`cli` v1 keys load as the right `(cli, scope)` tuple |
| `test_non_claude_scope_user_passes_through_to_global_config` | `--scope` is silently coerced for non-Claude CLIs; never produces a phantom `(cursor, project)` |

Two existing tests were updated to use the new `(cli, scope)` tuple keys (`test_codex_adds_block_when_absent_and_revert_removes_it`, `test_save_state_writes_atomically`).

### Docs

- `scripts/README.md` — new "`--scope {user,project}` (Claude only)" subsection with the two-layer config explanation, the silent-coerce-for-others rule, and the simultaneous-scopes paragraph. Safety section updated to note the scope-suffixed Claude backups and the v2 schema with v1 migration.
- `CHANGES` — entry under `### Development` (per project convention: dev tooling is not shipped product surface).

## Design decisions

- **Default `--scope project`**: keeps every existing call site — including `just mcp-use-local` from the libtmux-mcp repo itself — behaviourally identical. The flag is purely additive.
- **Silent coerce for non-Claude CLIs instead of erroring**: Codex / Cursor / Gemini have no per-project layer in their config files, so `--scope` is meaningless for them. Erroring would create friction in scripts that pass `--scope user` for cross-CLI consistency. `_normalize_scope` does the coerce in one spot.
- **Tuple keys (`(cli, scope)`) in memory; `cli:scope` strings on disk**: Tuple gives static guarantees on key shape; the string form is friendlier for the JSON file and stays human-editable.
- **v1 migration in `_parse_state_key`, not at write time**: `save_state` always emits v2; the migration runs once whenever a v1 file is read, then disappears on the next write. No long-lived "v1-mode" state to leak.
- **Scope in Claude backup suffix; non-Claude unchanged**: only Claude can produce same-second swap collisions (two scopes, one config file). Adding the suffix universally would break byte-compatibility with existing v1 `*.bak.mcp-swap-<ts>` files for users mid-revert.

## Verification

The flag-related code paths are all in `scripts/mcp_swap.py`:

```bash
rg -n 'scope' scripts/mcp_swap.py
```

State-file schema bump is reflected at one constant:

```bash
rg -n 'STATE_VERSION' scripts/mcp_swap.py
```

The two existing tests that reach into the in-memory state dict were updated; no other test file references the swap state shape:

```bash
rg -n '("claude"|"codex"|"cursor"|"gemini")\s*[:,].*Swap' tests/
```

## Test plan

- [x] `test_claude_user_scope_writes_top_level_mcpServers` — user-scope writes `mcpServers`, leaves `projects[<abs>]` untouched
- [x] `test_claude_user_scope_round_trip_restores_byte_identical` — round-trip preserves bytes
- [x] `test_claude_user_and_project_swaps_coexist_independently` — both scopes coexist; selective revert works
- [x] `test_legacy_v1_state_migrates_to_v2_keys` — v1 keys (`claude`, `codex`, ...) migrate to `(cli, scope)` tuples
- [x] `test_non_claude_scope_user_passes_through_to_global_config` — `--scope` silently coerced for non-Claude
- [x] All 18 existing `tests/test_mcp_swap.py` tests still pass (two updated for new key shape)
- [x] `just ruff` — clean
- [x] `just ruff-format` — no diff
- [x] `just mypy` — strict, zero errors
- [x] `uv run pytest --reruns 0` — full suite green
- [x] `rm -rf docs/_build && just build-docs` — same warning baseline as `main`

## Manual QA (real configs, post-backup)

Run from `/home/d/work/python/libtmux-mcp` after backing up `~/.claude.json` / `~/.codex/config.toml` / `~/.cursor/mcp.json` / `~/.gemini/settings.json` to a safe location:

```bash
just mcp-status                                                            # baseline shows [claude:user] PyPI + [claude:project] local
uv run scripts/mcp_swap.py use-local --cli claude --scope user --dry-run   # diff preview
uv run scripts/mcp_swap.py use-local --cli claude --scope user             # swap
(cd ~/work/cv && claude mcp list | grep libtmux)                           # → "uv --directory ... run libtmux-mcp"
uv run scripts/mcp_swap.py revert --cli claude --scope user                # revert just user-scope
(cd ~/work/cv && claude mcp list | grep libtmux)                           # → "uvx --no-config libtmux-mcp" (rolled back)
just mcp-status                                                            # project-scope still local
```

Confirmed end-to-end on this branch. The user-scope swap reaches every project that has no per-project override; reverting it leaves the per-project libtmux-mcp swap intact.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,9 +16,15 @@ _Notes on upcoming releases will be added here_
   Gemini have no per-project layer; the flag is silently coerced to
   `user` for them. A single install can hold both scopes
   simultaneously with independent backups, and `revert --scope user`
-  restores just one layer. State-file schema bumps to v2 (keys are
-  `cli:scope`) with transparent migration of v1 entries on load.
-  (#???)
+  restores just one layer. Full `revert` unwinds in LIFO order via an
+  explicit `seq_no` counter on each state entry — no reliance on dict
+  iteration through the JSON round-trip. Shape validation is
+  centralised in `_claude_user_servers` / `_claude_project_node`
+  helpers so read / write / delete fail symmetrically with
+  `RuntimeError` on a malformed Claude config. The state-file schema
+  is internal — no compatibility contract; running an older `mcp_swap`
+  against a newer `state.json` is undefined behaviour and not
+  supported. (#???)
 
 ### Documentation
 
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,22 @@
 _Notes on upcoming releases will be added here_
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### What's new
+
+**Dev scripts**
+
+- `scripts/mcp_swap.py` — `use-local` and `revert` accept
+  `--scope {user,project}` so a Claude install can target either the
+  top-level `mcpServers` fallback (every project without an override
+  picks it up) or the per-project `projects.<abs>.mcpServers` entry.
+  Default `project` preserves the previous behaviour. Codex / Cursor /
+  Gemini have no per-project layer; the flag is silently coerced to
+  `user` for them. A single install can hold both scopes
+  simultaneously with independent backups, and `revert --scope user`
+  restores just one layer. State-file schema bumps to v2 (keys are
+  `cli:scope`) with transparent migration of v1 entries on load.
+  (#???)
+
 ### Documentation
 
 - Bump gp-sphinx docs stack to v0.0.1a16 — docs site now renders
--- a/CHANGES
+++ b/CHANGES
@@ -6,9 +6,7 @@
 _Notes on upcoming releases will be added here_
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
-### What's new
-
-**Dev scripts**
+### Development
 
 - `scripts/mcp_swap.py` — `use-local` and `revert` accept
   `--scope {user,project}` so a Claude install can target either the
--- a/CHANGES
+++ b/CHANGES
@@ -24,7 +24,7 @@ _Notes on upcoming releases will be added here_
   `RuntimeError` on a malformed Claude config. The state-file schema
   is internal — no compatibility contract; running an older `mcp_swap`
   against a newer `state.json` is undefined behaviour and not
-  supported. (#???)
+  supported. (#35)
 
 ### Documentation
 
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -115,9 +115,9 @@ entries untouched.
   script writes.
 - **Custom binary install locations.** Detection is `shutil.which` plus
   the file existing at the configured global path. Homebrew, npm
-  prefixes (`~/.npm-global/bin`), and post-migration paths
-  (`~/.claude/local/claude`, `~/.gemini/local/gemini`) are picked up
-  only when the binary is already on `PATH`.
+  prefixes (`~/.npm-global/bin`), and the canonical local-install
+  layouts (`~/.claude/local/claude`, `~/.gemini/local/gemini`) are
+  picked up only when the binary is already on `PATH`.
 
 ### Extending to a new CLI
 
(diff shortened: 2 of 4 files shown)


tony:
### Code review

Found 1 issue:

1. `cmd_revert` without `--scope` for Claude with both scopes recorded restores in dict-insertion (i.e. swap-chronological) order rather than reverse. The two `atomic_write` calls target the same `~/.claude.json`; the second backup was snapshotted *after* the first swap, so it contains the first swap's changes. Iterating forward means the second write overwrites the first, leaving the file partially reverted (one scope's swap survives) while `state.json` is fully cleared. The new `test_claude_user_and_project_swaps_coexist_independently` only covers selective `--scope user` revert; the unscoped path is not exercised. Fix: iterate `cli_keys` in reverse-chronological order so each backup restores the prior state.

https://github.com/tmux-python/libtmux-mcp/blob/7c3e7e7548eca7ae70e1f06e26fcb54f1367bc0f/scripts/mcp_swap.py#L810-L827

PR #37, opened 2026-05-08
Make tmux MCP fire on bare 'pane' / 'window' / 'session'

tony:
## Summary

Bare *pane* / *window* / *session* prompts (without saying "tmux") were failing to trigger libtmux-mcp in Claude Code. This PR fixes that with three discrete server-side levers:

1. **`_INSTR_SCOPE` activation rule** — names positive triggers (pane, split, terminal, scrollback, copy mode, `%`/`@`/`$` IDs) and explicit anti-triggers (browser, VS Code, Cursor, Neovim, i3/sway/Hyprland, Jupyter cells, login/HTTP sessions). The full transmitted `instructions=` now fits Claude Code's 2KB budget across all three safety tiers.
2. **Discovery-anchor descriptions + `alwaysLoad` meta + `serverInfo.name="tmux"`** — six anchor docstrings (`list_panes`, `list_windows`, `list_sessions`, `snapshot_pane`, `search_panes`, `capture_pane`) carry `tmux` plus a buried synonym (terminal / shell / scrollback / multiplexer / workspace) that lands in BM25's search corpus. Three read-only anchors carry `anthropic/alwaysLoad: true` so a tiny tmux vocabulary stays visible without preloading every schema.
3. **HOOKS_GAP rationale restored** — *"Write-hooks survive process death"* is the load-bearing causal phrase that makes the read-only-hooks rule a correctness boundary instead of a soft preference. Paired with safety-tier paragraph compression so net byte cost is -1.

## Test plan

- [x] `uv run ruff check . --fix --show-fixes && uv run ruff format .`
- [x] `uv run mypy`
- [x] `uv run py.test --reruns 0 -vvv` (442 passed)
- [x] `just build-docs` (clean, 50 baseline warnings)
- [x] In-process budget probe: `readonly+TMUX_PANE` worst case is 2044 bytes (margin 4 of 2048)
- [ ] Manual probes in fresh Claude Code session: bare *"pane"*, *"current window"*, *"this session"* all route to `mcp__tmux__*`; bare *"browser tab"*, *"editor split"*, *"Jupyter cell"* do **not**

PR #42, opened 2026-05-08
tests: variable-length parametrize for 2KB-budget regression test

tony:
## Summary

The existing 2KB-budget regression test parametrized (3 tiers × 2 `tmux_pane` states), but every case used `%42` (3 chars) and `default` (7-char socket name). A user with a slightly longer custom socket name + a multi-digit pane id from a long session pushes the readonly worst case very close to 2048 bytes, and the static cases never exercised this — future text additions could silently put realistic runtime injections over the budget.

This PR:
- Collapses the two cross-product `@pytest.mark.parametrize` decorators into one explicit list so the variable-length stress case is a peer entry instead of expanding the cross-product space
- Adds `(TAG_READONLY, "%99", "/tmp/tmux-1000/dev-prod,12345,0")` as the 7th case. Exercises BOTH axes: multi-digit pane id and a longer-than-default socket name. Margin ~2 bytes from the 2048 ceiling.
- Inline comment names the fallback path (tighter compression form) if a future text addition trips this case.

## Note on base branch

This PR targets `pane-discoverability` (PR #37), not `main`. The variable-length stress test cannot pass off `main` alone — the 2KB compression in `pane-discoverability` is what makes the budget fit in the first place. Merge after #37 lands.

## Test plan

- [x] `uv run ruff check . && uv run mypy && uv run py.test -q` (443 passed; +1 stress case)
- [x] `just build-docs`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,7 +10,31 @@ import typing as t
 from gp_sphinx.config import make_linkcode_resolve, merge_sphinx_config
 from gp_sphinx.defaults import DEFAULT_SPHINX_FONT_PRELOAD
 
+# Docs-only shim: ``sphinx_autodoc_fastmcp.ToolCollector.tool()`` is a mock
+# of ``FastMCP.tool()`` and only accepts ``title``, ``annotations``, ``tags``.
+# Tools that pass ``meta=DISCOVERY_META`` for the per-tool
+# ``anthropic/alwaysLoad`` hint (verified at runtime via per-tool ``meta=``)
+# raise ``TypeError`` inside the mock, which silently drops the entire
+# enclosing module's tools from the docs catalog. Patch the mock to accept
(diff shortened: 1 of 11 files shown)


tony:
Folded into [#37](https://github.com/tmux-python/libtmux-mcp/pull/37) per the multi-model weave-ask synthesis (2-of-3 votes for fold; the variable-length parametrize is intrinsic to #37's 2KB-budget contract, not an independent feature).

The variable-length stress case (`%99` + `dev-prod` socket, margin ~2 bytes from 2048) now lives in the same commit that introduces `_INSTR_SCOPE` and the budget regression test on `pane-discoverability` — see [`4d5744e mcp(refactor[server]): Compress instructions, add SCOPE, fit under 2KB transmitted`](https://github.com/tmux-python/libtmux-mcp/commit/4d5744e).

Closing without merge.

PR #47, opened 2026-05-12
Pane(fix[wait_for_text]): waits for new output, joins wrapped lines, prefers wait-for-channel for completion

### LINKED ISSUE #45: `wait_for_text` matches stale scrollback immediately — needs baseline anchor
## Summary

`wait_for_text` does not capture a baseline before polling, so any text matching `pattern` that was already on the pane when the tool is invoked causes the very first poll (~50 ms later) to return `found=True` — regardless of whether the agent's command has produced any new output. The sibling tool `wait_for_content_change` already snapshots a baseline; `wait_for_text` doesn't, and so races against its own scrollback.

## Reproduction

```
1. Create a fresh pane.
2. Send: echo READY
3. Wait until the prompt returns.
4. Call wait_for_text(pane_id=…, pattern="READY", timeout=2).
   Expected: timeout, found=False (nothing new has been written).
   Actual:   found=True within ~50 ms, matching the line from step 2.
```

Relevant code paths:

- `src/libtmux_mcp/tools/pane_tools/wait.py` lines 100–254 — `wait_for_text`. The poll loop at lines 202–226 calls `pane.capture_pane(start=content_start, end=content_end)` on each tick and runs the compiled regex over every returned line. No state is recorded before the loop starts.
- `src/libtmux_mcp/tools/pane_tools/wait.py` lines 257–373 — `wait_for_content_change`. Line 327 (`initial_content = await asyncio.to_thread(pane.capture_pane)`) captures a baseline *before* the loop and compares each poll against it. This is the pattern `wait_for_text` should mirror.

## Why this matters for MCP DX

Agents call `wait_for_text` to synchronise on command output. A tool whose name reads as "wait" but whose behaviour reads as "match anywhere in the current pane" violates the principle of least astonishment, and there is no signal in the result that the match came from stale scrollback rather than fresh output. The synchronous "is the pattern currently in the pane?" intent is already served by `search_panes` (`src/libtmux_mcp/tools/pane_tools/search.py` lines 76–142), so `wait_for_text` should mean *wait for new appearance*.

### (A) Baseline-anchored, no flag — recommended

At entry, read `(history_size, cursor_y)` via `display-message` (the existing pattern used by `snapshot_pane` / `get_pane_info` in `src/libtmux_mcp/tools/pane_tools/meta.py` lines 128–162). The pair `(hs0, cy0)` defines an absolute grid index `baseline_abs = hs0 + cy0` that is stable across subsequent scrolling: tmux's `capture-pane -S <n>` interprets `n` against the live `hsize` (cmd-capture-pane.c lines 145–160, `top = gd->hsize + n`).

On each poll, read the current `history_size` (`hs1`) and capture only lines after the baseline: `pane.capture_pane(start=baseline_abs - hs1 + 1, end=None)`. The `+ 1` skips the baseline line itself so we don't re-match the line the cursor sat on at entry.

Why this works in practice: on the main screen, new content is monotonically appended at or near the cursor — `screen-write.c:screen_write_linefeed` is the only mover, and when it would advance past `sy-1` it scrolls the top line into history (`grid.c` `grid_scroll_history` increments `hsize`). Full-screen TUIs run on the alternate screen, which has a fresh grid with `hlimit=0` (`screen.c:screen_alternate_on`), so they don't interact with the baseline at all.

Pros: pure docstring + small implementation change, no new state to manage, no fd lifecycle, mirrors the in-tree pattern from `wait_for_content_change`. Breaking change in name only (returning `False` where today it returns `True` was the user's intent anyway), and pre-alpha (`0.1.0a0`) makes the cost negligible.

Cons: edge cases the docstring must call out — see the bottom of this issue.

### (B) Baseline-anchored, opt-in flag `skip_existing_matches=False`

Same mechanism as (A), but gated behind a boolean parameter that defaults to today's unsafe behaviour. Preserves callers that have somehow come to rely on stale-match semantics.

Pros: no breakage. Cons: every caller has to know to set a flag to get the only sensible default. MCP guidance (and FastMCP's own example tools, e.g. `examples/memory.py` splitting `remember()` and `read_profile()` instead of one `memory(mode=…)` tool) favours narrow tools over mode flags. The flag is effectively a permanent tombstone for a bug.

### (C) Sentinel-marker pattern, docs-only fix

Document that callers should send a unique marker (`cmd; echo __WAIT_$RANDOM__`) and wait for *that* marker. Works adversarially and addresses cases that even (A) can't fix (e.g. patterns the caller didn't originate, such as prompts or log lines).

Pros: no code change. Cons: pushes the cost onto every caller, doesn't help when the agent is waiting for output it didn't author (logs, third-party processes, prompts), and is easy to forget. Useful as a *complement* to (A), not a replacement.

### (D) `pipe-pane` side channel

Open a pipe-pane to a side channel at baseline time (`cmd-pipe-pane.c` lines 56–120), then read the pipe for new bytes. Strongest guarantee — sees only future bytes, immune to scrollback rollover, no polling.

Pros: most correct. Cons: heaviest — fd lifecycle, cleanup on pane death, a second reader loop, and changes the shape of the tool from "poll-and-return" to something stateful. Belongs in a future streaming tool (e.g. a `stream_pane`), tracked separately from this fix.

## Recommendation

Adopt **(A)**. The implementation is a small extension of `wait_for_text` that mirrors `wait_for_content_change`. The current `content_start` / `content_end` parameters can either be dropped (the baseline anchor supersedes them as a semantics knob) or repurposed as a performance cap only; (A) does not need them.

`search_panes` should be cross-referenced from the new docstring as the companion for "match anywhere right now." `(C)` should be mentioned in the docstring as an adversarial-safety pattern for callers who need stronger guarantees than tmux's grid model provides. `(D)` is deferred to a future tracking issue.

## Edge cases the docstring must call out

Regardless of which option is chosen, the new docstring should be explicit about:

- **Scrollback truncation.** If `history-limit` is small and the baseline line rolls out of history during the wait, `start` becomes more negative than `-hsize_now`. tmux clips `-S` to the oldest available line (`cmd-capture-pane.c` line 152), so the worst case degrades to today's behaviour on the surviving portion of history — not an infinite false-match loop.
- **Reverse-index sequences (`\eM`).** Programs that scroll content downward at the top of the scrolling region can rewrite history below the baseline. Rare on the main screen — pagers (`less`, `more`) and other heavy users run on the alternate screen and aren't observed via this tool.
- **`clear` / `reset`.** With the default `scroll-on-clear` option, cleared content scrolls into history (`screen-write.c:screen_write_clearscreen`), so baseline anchoring is unaffected.

## Out of scope

- `wait_for_content_change` — already anchored, no change needed.
- A streaming `pipe-pane`-based tool — track separately if added to the roadmap.
- `search_panes` semantics — already covers the synchronous case correctly.

tony:
Closes #45.

## What changed for users

`wait_for_text` becomes a "wait for output written after this call" primitive instead of "match anything currently in the pane." For agents whose first poll used to return `found=True` on stale scrollback, this is a real fix. The CHANGES `## libtmux-mcp 0.1.x (unreleased)` section describes the published-release-reader contract.

Concrete user-visible changes vs `0.1.0a6`:

- Anchors on `(history_size, cursor_y)` at entry; only matches lines written after the call begins.
- `content_start` / `content_end` parameters removed (the anchor supersedes them).
- Joins visually wrapped lines for matching via tmux's `-J` flag — long error strings that wrap across rows now match.
- Rejects empty `pattern`, sub-10 ms `interval`, non-positive `timeout` at entry.
- Raises `ToolError` on pane death, pane respawn, or `clear-history` mid-wait.
- `WaitForTextResult.timed_out` and `ContentChangeResult.timed_out` removed (mechanically `not found` / `not changed`).
- Emits a `notifications/message` warning when polling approaches `history-limit` — the band where tmux's `grid_collect_history` trim+scroll bounce keeps `hsize` clamped at the cap and the strict-shrink guard can't catch every trim.
- Server instructions, `send_keys` docstring, and `wait_for_text` docstring all point agents at `wait_for_channel` (composed with `tmux wait-for -S`) as the deterministic primitive for command completion. The polling-scraper `wait_for_text` is now framed for output the agent does not author.
- `docs/topics/prompting.md` system-prompt fragment and `docs/topics/troubleshooting.md` "Timing" bullet brought in line with the channel-first guidance.

## Theoretical grounding

Reviewed against tmux source pinned at `134ba6c`:

- Capture math: `cmd-capture-pane.c:158` (`top = gd->hsize + n`)
- Bottom-row clip: `cmd-capture-pane.c:159-160`
- Trim mechanics: `grid.c:386` (`grid_collect_history` decrements `hsize` by `hlimit/10`)
- Scroll-after-trim sequence: `grid-view.c:106` (`grid_view_scroll_region_up`)
- Resize-grow / hsize pull-back: `screen.c:451-465` (`screen_resize_y`)
- Pane respawn preserves `hsize`: `screen.c:127` (`screen_reinit`)
- No native trim-detection primitive exists: `format.c` history format strings + `options-table.c` hooks list both checked, no monotonic counter or `history-trimmed` hook

The rollover detection's partial coverage is acknowledged explicitly. A 2 ms high-frequency probe across ~3000 samples confirmed there is no sub-poll window where `hsize` dips below entry once entry was sub-cap — so the strict-shrink guard catches `clear-history` and the entry-at-cap edge but not bursty trim. The runtime `ctx.warning` in the trim-risk band is the agent-facing mitigation; `wait_for_channel` is the deterministic alternative.

## Test plan

```
rm -rf docs/_build; uv run ruff check . --fix --show-fixes; uv run ruff format .; uv run mypy; uv run py.test --reruns 0 -vvv; just build-docs
```

Green at every commit. Full test suite passes.

New regression tests cover: stale scrollback no-match (#45), bottom-row clip, mid-wait respawn, pane death under `remain-on-exit`, input rejection (empty pattern / tiny interval / non-positive timeout), wrap-spanning patterns, resize-grow not raising, history shrinking via `clear-history` raising, idle wait already in trim-risk band warning, burst wait advancing into trim-risk band warning, cancellation propagation, server instruction substring assertions, `send_keys` docstring cross-link.

## Filed follow-ups (deferred, not in this PR)

- tmux-python/libtmux-mcp#50 — `wait_for_text` two-call state/capture race within a single tick (architecture)
- tmux-python/libtmux-mcp#51 — `include_baseline_row=True` opt-in for same-row writes (enhancement)
- tmux-python/libtmux-mcp#52 — adaptive backoff to reduce subprocess pressure (performance)
- tmux-python/libtmux-mcp#53 — `wait_for_content_change` lifecycle parity with `wait_for_text` (enhancement)
- tmux-python/libtmux-mcp#54 — expose `risk_band_warned: bool` on `WaitForTextResult` for clients not subscribed to log notifications (enhancement)
- tmux-python/libtmux-mcp#55 — `search_panes` visual-row capture misses wrap-spanning patterns (parallel to `wait_for_text`'s `-J` fix; the slow path is a 1-line change, the fast path is structurally wrap-blind in tmux)
- tmux-python/libtmux#671 — investigate libtmux-side trim detection helper (broader infrastructure)

## Breaking changes

- `content_start` / `content_end` parameters removed from `wait_for_text`. Pre-alpha API; CHANGES carries a before/after block.
- `wait_for_text` raises `ToolError` on empty `pattern`, `interval < 0.01`, `timeout <= 0`, pane death, pane respawn, and `history_size` shrink below entry baseline. Previously most of these silently miscaptured or returned `found=False` after the full timeout.
- `WaitForTextResult.timed_out` and `ContentChangeResult.timed_out` removed. They were mechanically `not found` / `not changed`; callers should switch to the primary result field. CHANGES carries a before/after block.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -30,6 +30,10 @@ When the pane's `history-limit` causes the baseline grid row to be trimmed mid-w
 
 ### Fixes
 
+**{tooliconl}`wait-for-text` matches patterns across visually-wrapped lines**
+
+The poll-loop capture now passes tmux's `-J` flag (`join_wrapped=True` on `Pane.capture_pane`), so a pattern that crosses the pane's visual wrap is still matched against the joined logical line. Long error strings like `"Build failed: module not found"` that tmux split across two rows previously slipped through `re.search`. The joined logical line is returned in `matched_lines` and may exceed `pane_width`. (#45 follow-up)
+
 **{tooliconl}`wait-for-text` waits for new output, not stale scrollback**
 
 {tooliconl}`wait-for-text` now anchors on the pane's grid position at entry and only matches lines written after the call begins; the previous behaviour returned `found=True` on the first poll whenever the pattern already lived in the pane. Panes that die or are respawned mid-wait surface a `ToolError`. For the synchronous "is the pattern in the pane right now?" case, call {tooliconl}`search-panes`. (#45)
--- a/CHANGES
+++ b/CHANGES
@@ -20,9 +20,11 @@ wait_for_text(pattern="OK", content_start=-100)
 wait_for_text(pattern="OK")
 ```
 
-**{tooliconl}`wait-for-text` raises on history rollover**
+**{tooliconl}`wait-for-text` raises on a subset of history-rollover events**
 
-When the pane's `history-limit` causes the baseline grid row to be trimmed mid-wait (or `clear-history` runs), {tooliconl}`wait-for-text` now raises `ToolError` instead of silently returning stale matches or missing new output. The absolute baseline anchor relies on `history_size` not shrinking due to row eviction; once `grid_collect_history` frees those rows there is no server-side way to recover the original anchor. (Resize-grow also decrements `history_size`, but does not evict rows — that case is exempted by also checking `pane_height`.) For deterministic command-completion synchronization, compose `tmux wait-for -S <channel>` into the shell command and use {tooliconl}`wait-for-channel`. (#45 follow-up)
+When the pane's `history-limit` causes `hsize` to shrink below the entry baseline mid-wait — typically via `clear-history`, or via retroactive `history-limit` shrink on tmux >= 3.7 (commit `195a9cf`) — {tooliconl}`wait-for-text` now raises `ToolError` instead of silently returning stale matches or missing new output. Resize-grow also decrements `hsize` (rows pulled back from history), but does not evict row data; that case is exempted by also checking `pane_height`.
+
+The guard does **not** reliably detect `grid_collect_history` trim that occurs during continuous output: tmux's trim-then-scroll bounce keeps sampled `hsize` clamped near the cap, so the predicate stays false even though rows are being evicted. For deterministic command-completion synchronization, compose `tmux wait-for -S <channel>` into the shell command and use {tooliconl}`wait-for-channel`. {tooliconl}`wait-for-text` also emits a runtime `notifications/message` warning when polling within the trim-risk band (see ### Fixes), so MCP clients can surface "correctness is best-effort here" without the call failing. (#45 follow-up; tracking issue filed in libtmux for a deeper detection helper.)
 
 ### Documentation
 
--- a/CHANGES
+++ b/CHANGES
@@ -6,9 +6,33 @@
 _Notes on upcoming releases will be added here_
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### Breaking changes
+
+**{tooliconl}`wait-for-text` drops `content_start` / `content_end`**
+
+The baseline anchor introduced in this release follows the pane's grid position automatically, so the previous manual capture-range parameters have no remaining purpose. Agents that named them should drop them from their call sites. (#45)
+
+```python
+# Before
+wait_for_text(pattern="OK", content_start=-100)
+
+# After
+wait_for_text(pattern="OK")
+```
+
 ### Dependencies
 
 **Minimum `libtmux>=0.56.0`** (was `>=0.55.1`). Unlocks the new tmux-command wrappers shipped in libtmux 0.56.0 — {meth}`~libtmux.Pane.respawn`, {meth}`~libtmux.Pane.copy_mode`, {meth}`~libtmux.Pane.pipe`, {meth}`~libtmux.Pane.swap`, {meth}`~libtmux.Pane.paste_buffer`, {meth}`~libtmux.Pane.clear_history`, {meth}`~libtmux.Pane.display_message`, {meth}`~libtmux.Server.delete_buffer`, and the {meth}`~libtmux.Session.next_window` / {meth}`~libtmux.Session.previous_window` / {meth}`~libtmux.Session.last_window` trio — so the MCP no longer falls back to raw `cmd()` calls for tmux commands the upstream API now covers. (#46)
+
+### Fixes
+
+**{tooliconl}`wait-for-text` waits for new output, not stale scrollback**
+
+{tooliconl}`wait-for-text` now anchors on the pane's grid position at entry and only matches lines written after the call begins; the previous behaviour returned `found=True` on the first poll whenever the pattern already lived in the pane. Panes that die or are respawned mid-wait surface a `ToolError`. For the synchronous "is the pattern in the pane right now?" case, call {tooliconl}`search-panes`. (#45)
+
+**{tooliconl}`wait-for-text` rejects misused `pattern` / `interval` / `timeout`**
+
+Empty `pattern`, `interval` below 10 ms, and non-positive `timeout` each now raise `ToolError` at entry instead of silently matching every line, spinning the tmux server in a tight loop, or completing a surprise single probe. (#45)
 
 ## libtmux-mcp 0.1.0a6 (2026-05-09)
 
--- a/CHANGES
+++ b/CHANGES
@@ -50,7 +50,7 @@ Empty `pattern`, `interval` below 10 ms, and non-positive `timeout` each raise `
 
 **Wait family is re-framed around {tooliconl}`wait-for-channel` as the deterministic primitive**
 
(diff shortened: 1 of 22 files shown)


tony:
### Code review

tony:
### Code review

Found 1 issue:

1. CHANGES unreleased block orders the subheadings `### Breaking changes` → `### Fixes` → `### Dependencies`, but AGENTS.md mandates `### Breaking changes` → `### Dependencies` → `### What's new` → `### Fixes` → `### Documentation` → `### Development`. Move the Dependencies block to land between Breaking changes and Fixes.

https://github.com/tmux-python/libtmux-mcp/blob/fbe4698dd1c700a67da9c293316bf1f809d07a90/CHANGES#L7-L60

(AGENTS.md rule: https://github.com/tmux-python/libtmux-mcp/blob/fbe4698dd1c700a67da9c293316bf1f809d07a90/AGENTS.md#L419-L421)

tony:
### Code review

Found 2 issues:

1. CHANGES unreleased section ordering still violates AGENTS.md (the May 15 review's finding remains unresolved). AGENTS.md says "**Fixed subheadings**, in this order when present: `### Breaking changes`, `### Dependencies`, `### What's new`, `### Fixes`, `### Documentation`, `### Development`." Current order is `Breaking changes` → `Documentation` → `Dependencies` → `Fixes`. `### Documentation` must come last (after `### Fixes`), and `### Dependencies` must come before `### Fixes`.

https://github.com/tmux-python/libtmux-mcp/blob/91ea7448a00f79b9585d4c8e718a69ac3fa5cacd/CHANGES#L28-L44

AGENTS.md rule: https://github.com/tmux-python/libtmux-mcp/blob/91ea7448a00f79b9585d4c8e718a69ac3fa5cacd/AGENTS.md#L419-L421

2. `tests/test_pane_tools.py:2074` uses bash/zsh brace expansion `{1..100}` while every other pre-fill in the same file uses POSIX-portable `$(seq 1 N)` (8 other call sites: lines 1332, 1519, 1563, 1601, 1662, 1716, 2018). When the new pane's shell is `dash` or any non-bash POSIX `/bin/sh`, the literal string `{1..100}` is iterated once and produces no scrollback — the `_prefilled` `retry_until` then exhausts its budget before `wait_for_text` ever runs. Verified across all tmux 3.2a–master checkouts: shell selection sits below tmux, this is purely a shell-portability issue. Use `$(seq 1 100)` to match the file's convention.

https://github.com/tmux-python/libtmux-mcp/blob/91ea7448a00f79b9585d4c8e718a69ac3fa5cacd/tests/test_pane_tools.py#L2072-L2076

PR #57, opened 2026-05-17
feat(install widget): dependency-cooldown configurator

### LINKED ISSUE #31: [docs] Document how to bypass uv `exclude-newer` cooldowns when running `uvx libtmux-mcp`
## Problem

Users who configure a global `exclude-newer` in `~/.config/uv/uv.toml` (a popular supply-chain hardening pattern, e.g. `exclude-newer = "7 days"`) will silently miss new `libtmux-mcp` releases when running `uvx libtmux-mcp` or installing the server into Claude Code / Cursor / Codex / etc. Because libtmux-mcp is pre-alpha and shipping frequently, this is more than a theoretical concern — a fresh release can be invisible to anyone running with a 7-day or 30-day cooldown.

Today the docs say nothing about this. A repo-wide `rg 'exclude-newer|UV_EXCLUDE_NEWER|cooldown|UV_NO_CONFIG'` only matches `pyproject.toml`, which is developer-facing (the `[tool.uv.exclude-newer-package]` exemption block for `gp-*` workspace pins).

### 1. `docs/topics/troubleshooting.md` — primary

Add a new symptom-based section that matches the page's existing **Symptoms / Cause / Fix** pattern. Suggested heading: *"Old version of `libtmux-mcp` keeps running"*.

Content outline:

- **How to detect**: `uvx libtmux-mcp --version` shows an older release than [PyPI](https://pypi.org/project/libtmux-mcp/).
- **Root cause**: a global `exclude-newer` cooldown in the user's uv config filters newly-uploaded distributions out of the resolver.
- **Three escape hatches, in order of preference**:
  1. `UV_NO_CONFIG=1` in the MCP host's `env` block — skips uv config discovery for this server only. **Recommended** because it's portable across MCP hosts and documents intent.
  2. `--no-config` flag on the `uvx` invocation — same effect, requires editing `args` rather than `env`.
  3. `--exclude-newer 9999-01-01` (or `UV_EXCLUDE_NEWER=9999-01-01`) — overrides just the cutoff while leaving other uv config (indexes, cache dirs) intact. Useful if the user actually wants to keep most of their uv config.
- **Caveat**: the cooldown exists as a supply-chain defense — bypassing it for an MCP that auto-launches every Claude Code session reintroduces the very risk window the cooldown closes. Suggest `uv tool install libtmux-mcp` once (cooldown applies at install time) and pointing the MCP at the installed binary as a middle ground.

### 2. `docs/installation.md` — secondary

Short cross-reference under the "Run without installing" tab pointing at the new troubleshooting section. One sentence; full detail belongs in troubleshooting.

### 3. `docs/clients.md` — optional

Could add `UV_NO_CONFIG` to the example `env` blocks for users who want it baked into their MCP host config from the start. Deferrable.

## Acceptance

- [ ] Both pages render cleanly in the Sphinx build (`just build-docs`).
- [ ] The troubleshooting entry follows the existing **Symptoms / Cause / Fix** structure already used on the page.
- [ ] `CHANGES` gets a one-line entry under the "Documentation" subsection of the unreleased `0.1.x` block, consistent with the `0.1.0a2` doc-entry pattern.

## References

- uv settings reference: https://docs.astral.sh/uv/reference/settings/
- uv resolution concepts (per-package opt-out semantics): https://docs.astral.sh/uv/concepts/resolution/
- Related upstream discussion on global cooldowns: astral-sh/uv#16813, astral-sh/uv#18775

tony:
## Summary

- **Add** an interactive cooldown picker to the MCP install widget so readers can apply or bypass dependency cooldowns without leaving the page. Closes #31's docs request for a `uvx`-bypass path.
- **Extend** the install matrix to a fourth dimension (cooldown ∈ {off, days, bypass}) — 90 server-rendered panels covering every legal `(client, method, scope, cooldown)` cell.
- **Inject** the right per-tool flag automatically: uv's `--exclude-newer P<N>D` / `--no-config`, pip's `--uploaded-prior-to P<N>D` (pip 26.1+), pipx's `--pip-args=--uploaded-prior-to=P<N>D`. JSON/TOML config snippets gain `env: UV_NO_CONFIG=1` blocks for bypass.
- **Render** a per-panel italic caveat where bypass is a no-op (pip and pipx default backend have no global cooldown to skip).
- **Persist** mode + days globally to `localStorage` (`libtmux-mcp.mcp-install.cooldown.{mode,days}`); the prehydrate `<head>` script restores both before first paint so reloading never flashes through the default.
- **Fix** Sphinx 8's `copy_asset_file` no-overwrite behavior in `_assets.py` so widget JS/CSS actually re-ship on incremental builds (a quiet trap — the build reported success while serving stale assets).

### Data model & rendering

- **`docs/_ext/widgets/mcp_install.py`** — New `Cooldown` dataclass + `COOLDOWNS` tuple. `Panel` gains `cooldown`, `pip_prereq`, `note` fields. `_body_for(client, method, scope, cooldown)` returns `(body, language, note)`. `<DAYS>` sentinel marks the spot in days-mode bodies where the user-configurable day count slots in.
- **`docs/_ext/widgets/_base.py`** — New `cooldown_days_slot` Jinja filter swaps the post-Pygments `&lt;DAYS&gt;` for `<span data-cooldown-days-slot>7</span>` *inside* the highlighted string literal. The number inherits Pygments colors and stays selectable for copy-paste.
- **`docs/_ext/widgets/_prehydrate.py`** — Reads `cooldown.mode` / `cooldown.days` from localStorage in the inline `<head>` script. The `@layer mcp-install-prehydrate` rules enumerate every legal `(client × method × scope × cooldown)` quadruple so first paint always lands on the correct cell.
- **`docs/_ext/widgets/_assets.py`** — Replace `copy_asset_file` with `shutil.copy2` to force overwrites on incremental builds (Sphinx 8 emits `misc.copy_overwrite` warnings and aborts otherwise).

### Template & UX

- **`docs/_widgets/mcp-install/widget.html`** — Adds the right-edge cooldown control (checkbox + label button + `?` help) on the method-tab row, plus a parallel settings sub-view with radio mode picker, days input, and a collapsible *What are cooldowns?* explainer linking [cooldowns.dev](https://cooldowns.dev/) and the [Datadog Security Labs writeup](https://securitylabs.datadoghq.com/articles/dependency-cooldowns/).
- **`docs/_widgets/mcp-install/widget.css`** — Styles the cooldown control, settings view, days input, and the per-tool cooldown caveat row. CSS-only view swap via `html[data-mcp-install-view="settings"]` (no transitions — see commit `f23d7d5` for why).
- **`docs/_widgets/mcp-install/widget.js`** — `setCooldownMode()` / `setCooldownDays()` / `setView()` plus delegated handlers for the checkbox, radio, days input, back link, and help button. Auto-save on every change (matching the hsk-django flashcard settings pattern); `input` events update slot textContent in real time while `change` events persist + broadcast.

### Tests & changelog

- **`tests/docs/test_widgets.py`** — 4-axis cross-product asserts; per-tool body-shape tests for each cooldown mode; verifies the pip/pipx bypass caveat note; confirms `Panel.pip_prereq` is set only for the pip method and carries the cooldown flag when applicable.
- **`CHANGES`** — User-facing entry under `### What's new` per AGENTS.md's deliverable test.

## Design decisions

**Days slot via post-Pygments span injection.** Server-renders the snippet as `…P<DAYS>D…`. Pygments highlights the surrounding text, then the `cooldown_days_slot` filter swaps `&lt;DAYS&gt;` for a `<span data-cooldown-days-slot>7</span>`. The span lives inside the Pygments string-literal span so the number inherits its color; `widget.js` updates `textContent` on every input keystroke. Copy-paste captures the visible number. The alternative — CSS `::before { content: var(--days) }` — would have made the value uncopyable.

**View toggle is transient, mode is persistent.** Cooldown mode and days persist across pages and reloads. The settings-vs-install view does not — a user mid-form who reloads expects to see the snippet on next load, not the open form. Implementing view as `html[data-mcp-install-view]` (set only by JS, never restored by prehydrate) gets this for free.

**Bypass for non-uv tools is shown, not hidden.** Pip and pipx default backend have no global cooldown to bypass, so their bypass panels render the same body as off but surface an italic per-tool note. This is more honest than silently making the radio inert on certain method tabs, and the note links the user to the uv-backed fallback (`pipx[uv]`).

**Single source of truth for defaults.** `DEFAULT_COOLDOWN_MODE` / `DEFAULT_COOLDOWN_DAYS` live in Python and are serialized into the inline prehydrate script. Adding a new cooldown mode or moving the default later only requires editing the Python module — the script extends automatically.

## Verification

Every legal `(client, method, scope, cooldown)` quadruple has a matching CSS rule in the prehydrate `<style>` block. With the front page built, this should match 90:

```console
$ grep -oc 'data-mcp-install-cooldown-mode="[^"]*"\] \.lm-mcp-install__panel' docs/_build/html/index.html
```

The days slot span appears exactly once per days-mode panel (30 across the front page):

```console
$ grep -oc 'data-cooldown-days-slot' docs/_build/html/index.html
```

The bypass caveat note renders for the 20 panels where bypass is a no-op (pip + pipx, every scope row):

```console
$ grep -oc 'lm-mcp-install__cooldown-note' docs/_build/html/index.html
```

## Test plan

- [x] `uv run pytest --reruns 0` — widget unit tests cover 4-axis cross product, per-tool flag injection, pip/pipx bypass caveat notes, and `Panel.pip_prereq` being method-scoped.
- [x] `uv run ruff check .` + `uv run ruff format --check .`
- [x] `just mypy` — strict-typed widget framework passes.
- [x] `just build-docs` — all dirhtml targets build; the pre-existing duplicate-ID warnings in `docs/tools/**/*.md` are unrelated.
- [x] **Fresh load**: empty localStorage → front page shows the existing default snippet, cooldown checkbox unchecked, install view active.
- [x] **Toggle on**: click the checkbox → settings view opens, mode flips to `days` with the default 7, snippet shows `--exclude-newer P7D`.
- [x] **Real-time days update**: type `30` in the days input → all visible snippets update to `P30D` before blur (`input` event hook).
- [x] **Bypass mode**: pick *Bypass any global cooldown* → snippet swaps to `uvx --no-config libtmux-mcp` (CLI) or `env: { "UV_NO_CONFIG": "1" }` (JSON config).
- [x] **Per-tool caveat**: switch to the pip method while in bypass mode → small italic note appears below the snippet explaining pip has no global cooldown.
- [x] **Reload persistence**: configure bypass + days=14 → reload → first paint shows the bypass panel and the days input pre-filled to 14 with no visible flicker (prehydrate-driven).
- [x] **SPA navigation**: configure on `/`, navigate to `/clients/` → cooldown state survives the gp-sphinx soft-nav.
- [x] **Uncheck**: clicking the checkbox while a non-off mode is active resets mode to off and the snippets revert to their unflagged forms.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,10 @@ _Notes on upcoming releases will be added here_
 
 The MCP install widget on the front page, {doc}`/quickstart`, and {ref}`clients` now lets you pick a config scope per LLM client alongside the install method — *local* / *user* / *project* for Claude Code, *user* / *project* for Codex and Gemini, *project* / *global* for Cursor. JSON-only clients (Cursor, Claude Desktop) also show the destination config-file path next to the snippet so the paste target is never a guess. Your scope choice is remembered per-client, so switching between clients restores each one's last selection.
 
+**Installer picker can apply or bypass dependency cooldowns**
+
+A new *Configure cooldowns* control on the install widget lets you tack a cooldown onto the snippet without leaving the page. Pick a delay in days (uv's `--exclude-newer`, pip's `--uploaded-prior-to`, pipx's `--pip-args`) to wait out community vetting before pulling a fresh release, or pick *Bypass any global cooldown* to skip a `~/.config/uv/uv.toml` cutoff and grab the latest libtmux-mcp anyway. The setting persists across pages, and the embedded *What are cooldowns?* expander links to the [Datadog Security Labs writeup](https://securitylabs.datadoghq.com/articles/dependency-cooldowns/) and [cooldowns.dev](https://cooldowns.dev/) if you want the supply-chain context. (#31)
+
 ## libtmux-mcp 0.1.0a7 (2026-05-16)
 
 ### Breaking changes
--- a/docs/_ext/widgets/_assets.py
+++ b/docs/_ext/widgets/_assets.py
@@ -3,10 +3,10 @@
 from __future__ import annotations
 
 import pathlib
+import shutil
 import typing as t
 
 from sphinx.util import logging
-from sphinx.util.fileutil import copy_asset_file
 
 from ._base import BaseWidget
 
@@ -28,6 +28,13 @@ def install_widget_assets(
     every page includes them (same pattern as ``sphinx-copybutton``). This is
     intentionally simpler than per-page inclusion — the files are small and the
     docs are not bandwidth-constrained.
+
+    Uses :func:`shutil.copy2` directly. Recent Sphinx releases tightened
+    ``copy_asset_file`` to refuse overwriting (it emits a
+    ``misc.copy_overwrite`` warning and aborts), which leaves stale widget
+    JS/CSS on every incremental rebuild. The fix is to do the byte copy
+    ourselves: the cache-busting ``?v=<hash>`` querystring on the
+    ``add_*_file`` registration line keeps browser caches honest.
     """
     if app.builder.format != "html":
         return
@@ -47,5 +54,5 @@ def install_widget_assets(
             if not source.is_file():
                 continue
             dest.mkdir(parents=True, exist_ok=True)
-            copy_asset_file(str(source), str(dest))
+            shutil.copy2(str(source), str(dest / filename))
             register(f"{STATIC_SUBDIR}/{name}/{filename}")
--- a/docs/_ext/widgets/_base.py
+++ b/docs/_ext/widgets/_base.py
@@ -147,11 +147,11 @@ def make_cooldown_days_slot_filter() -> CooldownDaysSlotFilter:
     default_date = default_cooldown_date(DEFAULT_COOLDOWN_DAYS)
     duration_span = (
         '<span class="lm-mcp-install__cooldown-days"'
-        f' data-cooldown-duration-slot>P{DEFAULT_COOLDOWN_DAYS}D</span>'
+        f" data-cooldown-duration-slot>P{DEFAULT_COOLDOWN_DAYS}D</span>"
     )
     date_span = (
         '<span class="lm-mcp-install__cooldown-days"'
-        f' data-cooldown-date-slot>{default_date}</span>'
+        f" data-cooldown-date-slot>{default_date}</span>"
     )
 
     def _filter(html: object) -> markupsafe.Markup:
--- a/docs/_ext/widgets/_base.py
+++ b/docs/_ext/widgets/_base.py
@@ -117,30 +117,46 @@ class BaseWidget(abc.ABC):
 
 
 def make_cooldown_days_slot_filter() -> CooldownDaysSlotFilter:
-    """Return a Jinja filter that injects a days-slot ``<span>`` into HTML.
+    """Return a Jinja filter that injects a cooldown-date slot ``<span>``.
 
-    The Pygments highlighter escapes the ``<DAYS>`` sentinel emitted in
-    days-mode bodies (see :mod:`docs._ext.widgets.mcp_install`) to
-    ``&lt;DAYS&gt;``. This filter walks the highlighted output and
-    swaps every occurrence with a span whose default text content is
-    ``7``:
+    The Pygments highlighter escapes the ``<COOLDOWN_DATE>`` sentinel
+    emitted in days-mode bodies (see :mod:`docs._ext.widgets.mcp_install`)
+    to ``&lt;COOLDOWN_DATE&gt;``. This filter walks the highlighted
+    output and swaps every occurrence with a span whose default text
+    content is the ISO-date string for *today - default-days*:
 
     .. code-block:: html
 
-        <span class="lm-mcp-install__cooldown-days" data-cooldown-days-slot>7</span>
+        <span class="lm-mcp-install__cooldown-days"
+              data-cooldown-date-slot>2026-05-10</span>
+
+    We pick the *absolute date* form rather than ISO 8601 duration
+    (``P7D``) because pipx 1.8.0 bundles a pip older than 26.1; the
+    duration syntax fails there with ``Invalid isoformat``. Absolute
+    dates have worked since pip 26.0 and across uv's
+    ``--exclude-newer``, so the resulting snippet is portable across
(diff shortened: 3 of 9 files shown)


tony:
### Code review

Found 3 issues:

1. `widget.html` top docstring claims the `cooldown_days_slot` Jinja filter swaps the sentinel for a `<span data-cooldown-days-slot>`, but the slot was split into `data-cooldown-duration-slot` (uvx + pip) and `data-cooldown-date-slot` (pipx) in a later commit on the same branch. The comment's attribute name doesn't exist in the rendered HTML.

https://github.com/tmux-python/libtmux-mcp/blob/362a92bbbcc86c42463eaf57e6e33437e85eb361/docs/_widgets/mcp-install/widget.html#L18-L25

2. `widget.css` cooldown-days slot comment calls `data-cooldown-days-slot` the JS hook for updating textContent on settings save. The JS uses `[data-cooldown-duration-slot]` and `[data-cooldown-date-slot]` separately (see `widget.js` lines around 463 / 466) — `data-cooldown-days-slot` is not present anywhere in the codebase.

https://github.com/tmux-python/libtmux-mcp/blob/362a92bbbcc86c42463eaf57e6e33437e85eb361/docs/_widgets/mcp-install/widget.css#L195-L201

3. `mcp_install.py` module docstring and the `_DURATION_SENTINEL` / `_DATE_SENTINEL` comment block both name the filter `cooldown_slots`, but the registered Jinja filter (per [`_base.py#L108`](https://github.com/tmux-python/libtmux-mcp/blob/362a92bbbcc86c42463eaf57e6e33437e85eb361/docs/_ext/widgets/_base.py#L106-L110)) is `cooldown_days_slot`.

https://github.com/tmux-python/libtmux-mcp/blob/362a92bbbcc86c42463eaf57e6e33437e85eb361/docs/_ext/widgets/mcp_install.py#L22-L28

https://github.com/tmux-python/libtmux-mcp/blob/362a92bbbcc86c42463eaf57e6e33437e85eb361/docs/_ext/widgets/mcp_install.py#L231-L237

Issue #50, opened 2026-05-15
wait_for_text: two-call state/capture race within a single poll tick


tony:
**Type:** architecture · **Tier:** deferred · **Tool:** `wait_for_text`

## What's happening

Each poll in `wait_for_text` runs two tmux subprocess calls in sequence:

1. `_read_pane_state` issues `display-message` to read `history_size`, `cursor_y`, `pane_height`, `pane_pid`, `pane_dead`.
2. `pane.capture_pane(start=start_line, end=None, join_wrapped=True)` issues `capture-pane`, where `start_line = baseline_abs - state.history_size + 1`.

Between (1) and (2), tmux can scroll more lines into history. tmux's `capture-pane` computes `top = gd->hsize + n` against the **live** hsize at capture time ([cmd-capture-pane.c#L158](https://github.com/tmux/tmux/blob/134ba6c/cmd-capture-pane.c#L158)), not the hsize we sampled in step 1. So when N new rows scroll between the two calls:

- We pass `n = baseline_abs - hsize_at_step1 + 1`
- tmux computes `top = hsize_at_step2 + n = baseline_abs + 1 + (hsize_at_step2 - hsize_at_step1)`
- The captured window starts N rows past the row we wanted; those N rows are invisible to the wait this tick.

## When it matters

Single-tick latency under bursty output. The next poll usually picks the missed rows back up — unless the missed rows have already scrolled past the visible region and been collected by [`grid_collect_history`](https://github.com/tmux/tmux/blob/134ba6c/grid.c#L386), at which point the rollover guard fires and the wait raises. So the bug surface is:

- One-tick `interval` of latency on transient bursts (default 50 ms; bounded).
- Permanent miss only at the moment of history rollover — but rollover now raises.

In other words: the race exists but its impact is bounded by `interval` and capped at \"raise\" rather than \"silently wrong\" thanks to the rollover guard.

### 1. Re-read after capture, retry on drift

```python
state_pre = await asyncio.to_thread(_read_pane_state, pane)
start_line = baseline_abs - state_pre.history_size + 1
lines = await asyncio.to_thread(pane.capture_pane, start=start_line, ..., join_wrapped=True)
state_post = await asyncio.to_thread(_read_pane_state, pane)
delta = state_post.history_size - state_pre.history_size
if delta > 0:
    # capture started \`delta\` rows too late; re-issue with adjusted start
    ...
```

[...]

PR #78, opened 2026-06-13
Add send_keys_batch and clarify terminal workflow primitives

### LINKED ISSUE #49: Compose / bulk operations: run multiple actions in one MCP tool call (start with send_keys)
## Problem

Every MCP tool call costs at least one round-trip between the agent and the server. When an agent needs to send several keystroke sequences to one pane, or coordinate `send_keys` across multiple panes, it currently needs one MCP call per operation. That burns agent turns, MCP transport round-trips, and libtmux/tmux subprocess work for sequences that can be represented as a batch.

A composed surface should let an agent request "run these N send-key operations in this declared order, with this error policy" as one MCP call.

## tmux and FastMCP support the shape

tmux command queues already provide useful semantics for a future native optimization:

- Chained commands share a group via [`cmd_get_group()` assignment](https://github.com/tmux/tmux/blob/134ba6c/cmd-queue.c#L516).
- A command error removes subsequent commands in that group via [`cmdq_remove_group()`](https://github.com/tmux/tmux/blob/134ba6c/cmd-queue.c#L470) and the error path at [`cmd-queue.c#L780`](https://github.com/tmux/tmux/blob/134ba6c/cmd-queue.c#L780).
- `send-keys` accepts a single target pane through its `-t` target field, so cross-pane batching requires either multiple libtmux calls or a tmux command chain; see [`cmd-send-keys.c`](https://github.com/tmux/tmux/blob/134ba6c/cmd-send-keys.c#L33).
- tmux command grammar uses semicolon-separated commands; see [`cmd-parse.y`](https://github.com/tmux/tmux/blob/134ba6c/cmd-parse.y#L407).

FastMCP/Pydantic can expose the MCP input safely:

- `list[...]` parameters are normal in FastMCP examples, including nested Pydantic models in [`examples/complex_inputs.py`](https://github.com/PrefectHQ/fastmcp/blob/2bff372/examples/complex_inputs.py#L16).
- `Annotated[list[Op], Field(min_length=1, max_length=N)]` renders as ordinary JSON Schema bounds.
- Pydantic gives per-item validation locations, so malformed op #3 can be reported precisely.
- Discriminated unions are possible server-side, but FastMCP strips OpenAPI `discriminator` keys from client-facing schema in [`json_schema.py`](https://github.com/PrefectHQ/fastmcp/blob/2bff372/fastmcp_slim/fastmcp/utilities/json_schema.py#L77), so heterogeneous batches are less ergonomic for clients.

## Recommendation: start with `send_keys_batch`

Add a homogeneous batch tool first. Keep `send_keys` unchanged for the common single-operation case.

```python
class SendKeysOp(BaseModel):
    keys: str
    pane_id: str | None = None
    session_name: str | None = None
    session_id: str | None = None
    window_id: str | None = None
    enter: bool = True
    literal: bool = False
    suppress_history: bool = False


class SendKeysOpResult(BaseModel):
    pane_id: str | None
    success: bool
    elapsed_seconds: float
    error: str | None = None


async def send_keys_batch(
    operations: Annotated[list[SendKeysOp], Field(min_length=1, max_length=50)],
    on_error: Literal["stop", "continue"] = "stop",
    socket_name: str | None = None,
    ctx: Context | None = None,
) -> list[SendKeysOpResult]: ...
```

Default `on_error="stop"` should short-circuit after the first failed operation. `on_error="continue"` should record the failure for that item and proceed with later operations in declared order.

## Deferred avenues

- Heterogeneous pane-op batches such as `send_keys -> wait_for_text -> capture_pane` are intentionally deferred. They have discovery/schema costs and overlap with the observation-first work in #60/#61.
- Native tmux command chaining is a future optimization. Start with libtmux calls in declared order because per-item error attribution is clearer.
- Progress notifications can be emitted for longer batches, but small batches should not depend on notification delivery for correctness.

## Acceptance criteria

- New `send_keys_batch` tool with Pydantic-validated `operations` and explicit `on_error` policy.
- Per-operation result list preserves declared order and includes success/error/timing metadata.
- Tests cover input bounds, declared ordering, `on_error="stop"`, `on_error="continue"`, cross-pane operations, and validation error reporting.
- Docs explain when to use `send_keys_batch` versus single `send_keys`, and call out that command completion should still use `wait_for_channel` when the agent authors the command.
- CHANGES describes the user-facing capability as one batch send-keys surface, not as internal tmux mechanics.

### LINKED ISSUE #61: Observation-first architecture: eliminate agent lockout from blocking wait tools
## 1. Problems

Agents get trapped when a pane wait hides the terminal behind a boolean. In a live dev loop, the agent is often tailing a server, watching tests, checking several panes, or deciding whether to interrupt. A blocking `wait_for_text` call gives it only `found` or timeout, so the agent cannot see that the server picked a different port, printed an error, is waiting for input, already finished, or that another pane now needs attention. This is the lockout reported in #59.

The deeper problem is tool mismatch. `wait_for_text` is useful for known future output from a process the agent does not control. It is a poor default for command completion, current-screen inspection, fuzzy diagnosis, or multi-pane monitoring. Those cases need different primitives:

| Agent intent | Better primitive | Common trap |
|---|---|---|
| "Did the command I sent finish?" | `wait_for_channel` with `tmux wait-for -S` | Waiting for a prompt or regex in `wait_for_text` |
| "What is visible right now?" | `snapshot_pane` / `capture_pane` | Calling a future-only wait |
| "Which pane mentions this?" | `search_panes` | Listing pane metadata and missing terminal text |
| "What changed since I last looked?" | Proposed `capture_since` (#60) | Repeated full snapshots or a long blocking wait |
| "Did this exact third-party line appear?" | `wait_for_text` | Treating it as a generic completion primitive |
| "Did anything change?" | `wait_for_content_change`, then inspect | Treating `changed=True` as semantic success |

The filed wait-family issues are concrete instances of that mismatch:

- #45 fixed stale scrollback instant matches by redefining `wait_for_text` as "new text after the call begins." That makes `search_panes` / `snapshot_pane` the correct answer for "is this already on screen?"
- #59 covers the 90-second agent lockout when the expected pattern is wrong, misspelled, already scrolled away, or semantically too narrow.
- #51 covers same-row output: `printf`, carriage-return progress bars, spinners, and status redraws can stay on the baseline row that `wait_for_text` deliberately skips to avoid stale matches.
- #55 covers wrap-spanning search misses: tmux's [`window_pane_search()`](https://github.com/tmux/tmux/blob/134ba6c/window.c#L1320) searches visual rows, so long build/test messages can split across the wrap boundary and silently miss on the fast path.
- #50 covers the two-call state/capture race: `wait_for_text` samples pane state, then calls `capture-pane`; tmux capture math is evaluated against live history in [`cmd-capture-pane`](https://github.com/tmux/tmux/blob/134ba6c/cmd-capture-pane.c#L107), while history can move through [`grid_collect_history()`](https://github.com/tmux/tmux/blob/134ba6c/grid.c#L386).
- #52 covers subprocess pressure under parallel waits: a 50 ms poll interval can become many tmux subprocesses per second across panes and agents.
- #53 covers lifecycle blindness in `wait_for_content_change`: pane death, respawn, and `pane_pid` changes are observation facts, not just content-diff edge cases.
- #54 covers notification-only risk signals: `ctx.warning()` may not reach the model or programmatic caller, so best-effort correctness must be exposed in structured tool results.
- #18 fixed the server-side version of this class for `wait_for_channel`: blocking `subprocess.run()` on the FastMCP event loop could stall the server. That is separate from agent lockout, where the server may stay healthy but the agent is still trapped inside an uninformative call.

The upstream substrates matter:

- tmux `wait-for` is the right zero-poll primitive for authored command synchronization; see [`cmd-wait-for.c`](https://github.com/tmux/tmux/blob/134ba6c/cmd-wait-for.c#L121).
- tmux grid/capture APIs are snapshots, not a durable log stream. History rollover, resize, alternate screen, and wrap behavior all affect what can be reconstructed from [`capture-pane`](https://github.com/tmux/tmux/blob/134ba6c/cmd-capture-pane.c#L107), [`grid_collect_history()`](https://github.com/tmux/tmux/blob/134ba6c/grid.c#L386), and alternate-screen handling in [`screen.c`](https://github.com/tmux/tmux/blob/134ba6c/screen.c#L676).
- FastMCP can report progress through [`Context.report_progress()`](https://github.com/PrefectHQ/fastmcp/blob/2bff372/fastmcp_slim/fastmcp/server/context.py#L389), and background tasks can send task status notifications, but UI/client notifications are not a substitute for agent-visible tool results.
- Python cancellation and timeouts need explicit async subprocess care. CPython documents `asyncio.create_subprocess_exec()` and `asyncio.wait_for()` as the async path, while `subprocess.run(timeout=...)` kills only the child it is managing after timeout; see [`asyncio-subprocess.rst`](https://github.com/python/cpython/blob/2b6a137/Doc/library/asyncio-subprocess.rst#L165) and [`subprocess.py`](https://github.com/python/cpython/blob/2b6a137/Lib/subprocess.py#L642).
- Shell composition is portable only at the simple `cmd; tmux wait-for -S channel` level. zsh documents list/separator and status behavior in [`grammar.yo`](https://github.com/zsh-users/zsh/blob/3e72a52/Doc/Zsh/grammar.yo#L66) and [`params.yo`](https://github.com/zsh-users/zsh/blob/3e72a52/Doc/Zsh/params.yo#L719). fish has different status variables, events, and prompt markers; see [`fish_for_bash_users.rst`](https://github.com/fish-shell/fish-shell/blob/fb29c85/doc_src/fish_for_bash_users.rst#L176), [`language.rst` events](https://github.com/fish-shell/fish-shell/blob/fb29c85/doc_src/language.rst#L2108), and [`mark-prompt`](https://github.com/fish-shell/fish-shell/blob/fb29c85/doc_src/language.rst#L2058). Status-preserving recipes should be shell-specific, not implied as universal.

## 2. Avenues of investigation

Primary investment: build `capture_since` (#60) as the observation-first primitive. It should be non-blocking, cursor-based, and bounded by `max_lines` / `max_bytes`, returning actual pane output plus cursor and correctness metadata. The first call should be immediately useful by returning visible screen content and establishing a cursor; later calls should return only deltas. This lets the agent tail dev servers, watch tests, inspect multiple panes, and self-correct without betting on one regex.

Keep `wait_for_channel` as the deterministic authored-command primitive. When the agent sends the command, compose a signal into the command text and wait on tmux's channel:

```console
$ command; tmux wait-for -S libtmux_mcp_done
```

That path is zero-poll inside tmux and avoids the prompt/regex/stale-output trap. The docs should keep the simple recipe portable, then add shell-specific status-preserving variants only after they are verified for zsh, fish, and POSIX-like shells.

Investigate FastMCP `task=True` as an optional enhancement, not the foundation. FastMCP supports task-enabled async tools through [`TaskConfig`](https://github.com/PrefectHQ/fastmcp/blob/2bff372/fastmcp_slim/fastmcp/utilities/tasks.py#L35), server-side task docs ([`docs/servers/tasks.mdx`](https://github.com/PrefectHQ/fastmcp/blob/2bff372/docs/servers/tasks.mdx#L49)), and client-side `task=True` calls ([`docs/clients/tasks.mdx`](https://github.com/PrefectHQ/fastmcp/blob/2bff372/docs/clients/tasks.mdx#L19)). This may reduce client/server lockout for long operations, but it does not by itself solve the core problem: the agent still needs model-visible pane output while deciding what to do.

Investigate event-driven pane streams separately. tmux has real streaming surfaces: [`pipe-pane`](https://github.com/tmux/tmux/blob/134ba6c/cmd-pipe-pane.c#L43) and control-mode output via [`control_write_output()`](https://github.com/tmux/tmux/blob/134ba6c/control.c#L474). These may support a future `stream_pane`, but they introduce lifecycle management, cleanup, raw PTY bytes, ANSI/control-sequence handling, buffering, safety-tier concerns, and stateful subscriptions. They should not block the simpler `capture_since` path.

Tighten the existing wait/search tools while the observation primitive is explored. #50-#55 still reduce real traps: race accounting, same-row behavior, adaptive backoff, lifecycle parity, structured risk fields, and wrap-aware search fallback all improve the current surface even if the recommended workflow shifts away from blocking waits.

Rewrite docs and server instructions around scenarios rather than tool names:

| Scenario | Recommended path |
|---|---|
| I launched a command and need completion | `send_keys("cmd; tmux wait-for -S chan")` + `wait_for_channel("chan")` |
| I need to inspect an already-running pane | `snapshot_pane` / `capture_pane` first |
| I need to tail or monitor output | `capture_since` once available |
| I need to search visible/current pane text | `search_panes`; use slow path where wrap-spanning text matters |
| I need a known future line from third-party output | `wait_for_text` with bounded timeout |
| I only need to know something changed | `wait_for_content_change`, followed by inspection |

Recommended ordering:

1. Ship documentation improvements immediately so agents stop reaching for `wait_for_text` as a generic completion tool.
2. Implement `capture_since` as the main new primitive (#60).
3. Continue the targeted correctness tickets (#50-#55) where they remain valuable.
4. Add optional FastMCP task support only after client behavior and dependency cost are understood.
5. Treat `pipe-pane` / control-mode streaming as a separate future `stream_pane` design, not a retrofit of blocking wait tools.

This keeps the architecture simple: make observation cheap, explicit, and model-visible; reserve blocking waits for cases where blocking is actually the right abstraction.

tony:
## Summary

- **Add `send_keys_batch`** — sends an ordered batch of raw key/text operations to tmux panes and returns a typed per-operation result (index, resolved pane, success, error, elapsed). `on_error` selects stop-at-first-failure or continue-and-report, an optional `timeout` aborts a long-running batch, and every operation carries its own pane/window targeting.
- **Keep the batch homogeneous by design** — it sends raw input only and does not compose `send → wait → capture`. Authored command completion stays with `run_command`; repeated observation stays with `capture_since`. This preserves clean per-operation error attribution instead of growing a workflow DSL inside one tool.
- **Harden audit redaction for batched payloads** — nested `operations[].keys` are digested to a length + SHA-256 prefix exactly like top-level `send_keys`, and malformed or schema-violating operation payloads are redacted by shape before the audit log can persist a raw value from an invalid request.
- **Preserve recovery suggestions in batch errors** — each per-operation error string now carries the `suggestion` hint from the underlying `ToolError`, so a client recovering from one failed operation gets the same guidance a single `send_keys` call would.
- **Reframe the I/O guidance around three primitives** — docs, prompt recipes, and tool docstrings now route authored shell commands through `run_command`, raw TUI / persistent-shell input through `send_keys` / `send_keys_batch`, and repeated observation through `capture_since`, with `wait_for_channel` reserved for custom shell completion.

### Tool surface
- **`src/libtmux_mcp/tools/pane_tools/io.py`**: `send_keys_batch` runtime — ordered iteration, `on_error` handling, a cooperative `timeout` checked between operations, and suggestion-preserving error capture.
- **`src/libtmux_mcp/models.py`**: `SendKeysOperation` (strict, `extra="forbid"`), `SendKeysOperationResult`, and `SendKeysBatchResult`.
- **`src/libtmux_mcp/tools/pane_tools/__init__.py`**, **`server.py`**: register the tool at the mutating tier and refresh the server instructions.

### Audit middleware
- **`src/libtmux_mcp/middleware.py`**: digest nested `operations[].keys`; redact malformed, non-list, and unknown-field operation payloads by shape before they can be logged.

### Prompts & docs
- **`src/libtmux_mcp/prompts/recipes.py`**: rewrite `run_and_wait` to teach `run_command` instead of a manual `send_keys` + `wait_for_channel` + `capture_pane` sequence.
- **`docs/`**: new `send-keys-batch` tool page; prompts, quickstart, recipes, gotchas, and topic guides reframed around the three primitives; Sphinx cross-reference roles fixed for the new tools.
- **`README.md`**, **`CHANGES`**: workflow overview and unreleased changelog entry.

### Tests
- **`tests/test_pane_tools.py`**: ordering, stop/continue, empty rejection, unknown-field rejection, timeout abort, suggestion preservation, and docstring routing.
- **`tests/test_middleware.py`**: redaction of well-formed, malformed, and nonschema batch payloads.
- **`tests/test_prompts.py`**, **`tests/test_server.py`**: updated prompt template and server-instruction expectations.

## Design decisions

- **Homogeneous batch, not a workflow DSL**: `send_keys_batch` deliberately refuses heterogeneous `send → wait → capture` steps. Mixing operation kinds would blur per-operation error attribution and grow an ad-hoc orchestration language inside a single MCP tool; authored completion already has `run_command` and observation has `capture_since`.
- **Redaction runs before schema validation**: the audit middleware summarizes raw arguments before fastmcp validates them, so it cannot assume well-formed `SendKeysOperation` dicts. It checks each field by type and redacts anything unexpected — a bare string, a non-dict list item, or an unknown key can never leak a payload into the log.
- **Cooperative timeout, not preemptive**: `timeout` is evaluated between operations rather than interrupting an in-flight `send_keys`. tmux key delivery is effectively instantaneous, so the guard exists to bound a long batch and stop spending agent tokens — not to kill a single hung send.

## Test plan

- [x] Full verification chain: `uv run ruff check .`, `uv run ruff format .`, `uv run mypy .`, `uv run py.test --reruns 0`, `just build-docs`
- [x] `test_send_keys_batch_sends_operations_in_order` — operations dispatch in submitted order with per-operation results
- [x] `test_send_keys_batch_stops_after_operation_error` / `test_send_keys_batch_continues_after_operation_error` — `on_error` semantics and `stopped_at`
- [x] `test_send_keys_batch_rejects_empty_operations` — empty batch raises an expected error
- [x] `test_send_keys_operation_rejects_unknown_fields` — `extra="forbid"` rejects misnamed fields
- [x] Batch `timeout` abort and per-operation suggestion preservation
- [x] `test_summarize_args_redacts_send_keys_batch_operations` plus malformed / nonschema payload redaction
- [x] `test_send_keys_docstring_routes_authored_commands_to_run_command` — docstring routes authored commands to `run_command`

Closes #49.
Closes #61.
Refs #50.


tony:
### Code review

tony:
### Code review

Re-reviewed the commits added since the previous review: the `send_keys_batch` `timeout` parameter, error-suggestion preservation, and the audit fix that redacts schema-validation inputs from FastMCP logs and tool error results before they render. The redaction was verified to close the leak on both paths (`_format_schema_validation_error` and the `fastmcp.server.server` log filter), with the strict-xfail reproduction promoted to a passing regression. No issues found. Checked for bugs and AGENTS.md compliance.

PR #81, opened 2026-06-20
Typed tmux operation chains

tony:
## Summary

- **Replace** the raw `run_command_chain` and narrow `build_forward_layout` tools with one typed `run_tmux_operations` tool.
- **Add** discriminated Pydantic operation models for `split_pane`, `send_keys`, `resize_pane`, `select_layout`, `set_option`, and `capture_pane`, plus structured per-step and per-dispatch results.
- **Compile** ordered typed operations to the fewest safe native tmux dispatches: no-output mutations fold into `tmux a ; b ; c`, output/id capture steps stay attributable, and `on_error="continue"` uses standalone dispatches because native tmux chains abort on first failure.
- **Preserve** the single-split decoration fast path: an id-producing `split_pane` can fold with immediate `send_keys` / `resize_pane` operations that target its `pane_ref` through tmux's `{marked}` target while still returning the concrete pane id.
- **Pin** the libtmux dependency to a public commit instead of a sibling worktree path.

### `src/libtmux_mcp/tools/chain_tools.py`

Adds the typed compiler and registers `run_tmux_operations` as a mutating, open-world tool. The compiler keeps a pending chain of no-output operations, flushes before output reads, captures split ids when needed, and returns both rendered argv and per-step status.

### `src/libtmux_mcp/models.py`

Adds the typed operation union and structured result models. The operation list is validated through a module-level Pydantic `TypeAdapter` in the tool module.

### Docs

Adds chain-tool docs and wires `chain_tools` plus the new models into the FastMCP docs catalog.

### Dependency pin

Updates `[tool.uv.sources]` and `uv.lock` to use the public libtmux branch commit `591a312f78d165816bb95a035a46219657c4b53d`.

## Design notes

- The tool is **typed**, not raw tmux argv. Unsupported commands such as `kill-server` are not operation variants.
- Per-step stdout is only promised where it can be attributed. `capture_pane` and id-producing split steps force a dispatch boundary unless the single split-ref `{marked}` optimization applies.
- `on_error="continue"` deliberately disables native chaining. tmux `;` sequences have abort-on-first-error semantics, so continuing later operations requires separate dispatches.
- The true one-connection, per-command-result path remains a future control-mode runner; this PR stays on the subprocess chain API.

## Test plan

- [x] `rm -rf docs/_build`
- [x] `uv run ruff check . --fix --show-fixes`
- [x] `uv run ruff format .`
- [x] `uv run mypy .`
- [x] `uv run py.test --reruns 0 -vvv` (`612 passed`)
- [x] `just build-docs`

## Companion PR

Depends on the libtmux experimental chain API in tmux-python/libtmux#685.


tony:
### Code review

Found 4 issues:

1. `ResizePaneOperation(zoom=False)` toggles zoom instead of leaving it unchanged: `_resize_pane_calls` appends `-Z` whenever `zoom is not None`, but `False is not None`, and `tmux resize-pane -Z` only toggles (there is no zoom-off), so `zoom=False` flips zoom — the opposite of its apparent meaning. The model validator accepts `zoom=False`.

https://github.com/tmux-python/libtmux-mcp/blob/55ea0b42624fdfeaaa66d978f797fa1343258420/src/libtmux_mcp/tools/chain_tools.py#L200-L205

2. `SetOptionOperation` validates target-without-scope but not scope-without-target, so `scope="pane"`/`"window"`/`"session"` with `target=None` passes validation and emits e.g. `set-option -p <opt> <val>` with no `-t`, silently mutating the control client's current pane/session.

https://github.com/tmux-python/libtmux-mcp/blob/55ea0b42624fdfeaaa66d978f797fa1343258420/src/libtmux_mcp/models.py#L796-L803

3. `ControlModeRunner` (the default `transport="control"`) is closed only on the normal-exit path, not in a `try/finally`. An exception in the dispatch loop (e.g. `asyncio.CancelledError`, or anything re-raised by `handle_tool_errors_async`) skips `close()` and leaks the `tmux -C` subprocess (it has no `__del__`); these accumulate over the server's lifetime.

https://github.com/tmux-python/libtmux-mcp/blob/55ea0b42624fdfeaaa66d978f797fa1343258420/src/libtmux_mcp/tools/chain_tools.py#L1088-L1093

4. CHANGES describes an intermediate branch state for a never-released tool ([AGENTS.md](https://github.com/tmux-python/libtmux-mcp/blob/55ea0b42624fdfeaaa66d978f797fa1343258420/AGENTS.md#L508-L511) says "Do not mention intermediate branch states, abandoned approaches, or 'no longer' behavior unless users of a published release actually experienced the old state"). `run-tmux-operations` is new in this PR, so "now dispatches ... by default" and "a failing operation no longer aborts the rest" describe a transition no released user experienced.

https://github.com/tmux-python/libtmux-mcp/blob/55ea0b42624fdfeaaa66d978f797fa1343258420/CHANGES#L23-L30

tony:
**Automated review finding.** Recorded here so it is not lost. Nothing has been pushed to this branch and no fix is applied — this is a note, not a change request.

### `rollback_on_error` bypasses the destructive-tier gate

In `src/libtmux_mcp/tools/chain_tools.py`, `run_tmux_plan`'s rollback path calls `kill-pane` directly, without the safety-tier check that gates `kill_pane` everywhere else.

Failure scenario: a server started at the `mutating` tier runs a plan with `rollback_on_error=true`; the plan fails partway, and rollback kills a pane anyway — the exact destructive action the tool's own gate says requires the `destructive` tier. The gate is bypassed precisely when something has already gone wrong, which is the worst moment to lose it.

Suggested fix: route the rollback's `kill-pane` through the same tier check as the user-facing path, and fail the rollback loudly if the tier does not permit it, rather than silently performing it.

PR #98, opened 2026-07-12
Do not retry a stale id; say how to fix an ambiguous target

### LINKED ISSUE #97: Grouped and linked windows crash every pane/window-targeted tool: MultipleObjectsReturned escapes the error mapper
Every tool that resolves a `pane_id` or `window_id` raises an opaque `Internal error` against a **grouped session** or a **linked window**. Grouped sessions are ordinary usage — `tmux new-session -t <existing>` — and are what tmuxp creates.

`pane_id` is the identifier the server instructions explicitly tell agents to *prefer*.

## Reproduction

```python
s = Server(socket_name="probe")
a = s.new_session(session_name="alpha")
s.cmd("new-session", "-d", "-t", a.session_name, "-s", "grouped")   # ordinary grouped session

s.panes.get(pane_id="%0", default=None)
```

```
list-panes -a rows for %0 -> [('%0', '$0'), ('%0', '$1')]

  server.panes.get(pane_id=%0)     -> RAISED MultipleObjectsReturned  (msg='')
  server.windows.get(window_id=@0) -> RAISED MultipleObjectsReturned  (msg='')
     isinstance LibTmuxException? False
```

## Why it escapes everything

`tmux list-panes -a` / `list-windows -a` emit **one row per holding session**, so a grouped or linked window matches twice. `QueryList.get()` raises `MultipleObjectsReturned` on more than one match **before** it consults `default`, so `default=None` does not absorb it.

`MultipleObjectsReturned` subclasses bare `Exception`, **not** `LibTmuxException`. So it misses every arm of `_map_exception_to_tool_error` and lands in the catch-all, surfacing as `ToolError("Unexpected error: MultipleObjectsReturned: ")` — with an **empty message**, `expected=False`, logged at ERROR.

`rg MultipleObjectsReturned` across `src/` and `tests/`: **zero hits.** Never caught, never tested.

## Blast radius

Both central resolvers:

- `_resolve_pane` — `_utils.py:716`, `server.panes.get(pane_id=..., default=None)`
- `_resolve_window` — `_utils.py:646`, `server.windows.get(window_id=..., default=None)`

which means `send_keys`, `capture_pane`, `get_pane_info`, `get_window_info`, `select_window`, `rename_window`, `move_window`, `split_window`, `wait_for_text`, and others.

The self-kill guards are on the same path — `session_tools.py:264` and `window_tools.py:331` both do `server.panes.get(pane_id=caller.pane_id, default=None)` — so the guard raises instead of guarding.

`list_windows` also returns duplicate rows for a linked window.

## Fix

Route the resolvers through libtmux's point lookups instead of filtering a server-wide listing:

```python
Pane.from_pane_id(server=server, pane_id=pane_id)
Window.from_window_id(server=server, window_id=window_id)
```

These target the object with `-t` and let tmux's own `cmd_find` resolve it, so exactly one row comes back.

**Sequencing matters.** On the pinned `libtmux>=0.61.0` those constructors still scan the whole server with `-a` and keep the *last* matching row — ordered by session **name** — so migrating today would trade a crash for a silently arbitrary session (the session you get would change if someone renamed a session). tmux-python/libtmux#713 rewrites them to `("-t", <id>)`. Do the migration **after** that lands and you get tmux's own answer; do it before and you inherit the arbitrary one.

So: this is not blocked on libtmux for the *crash* (you could catch `MultipleObjectsReturned` today), but the *correct* fix wants libtmux#713 released first.

## Interim mitigation

If a fix is wanted before that release, `MultipleObjectsReturned` at minimum needs an arm in `_map_exception_to_tool_error` so the agent gets an actionable message rather than an empty `Internal error`.

tony:
**Draft — do not merge with the git pin in place.** Validates [tmux-python/libtmux#718](https://github.com/tmux-python/libtmux/pull/718) downstream before it ships. Drop the `[tool.uv.sources]` stanza and bump the `libtmux` floor once #718 is released.

Fixes #97 (the ambiguous-lookup half).

## Why

libtmux#718 moves `MultipleObjectsReturned` and `ObjectDoesNotExist` under `LibTmuxException` — and `TmuxObjectDoesNotExist` inherits from `ObjectDoesNotExist`, so it moves too. This server keys two behaviours off `LibTmuxException`, and both change silently.

**The suite did not notice.** Pinned to the #718 branch, `main` is green: **701 passed, 0 failed.** Every changed path was uncovered. That is the finding, and it is why this PR leads with tests.

## What was broken

**1. A stale id was retried.** `ReadonlyRetryMiddleware` retries on `LibTmuxException`, which is correct for the transient socket errors libtmux wraps — but most of that family is not transient. A pane that does not exist will not exist on the second look; an ambiguous match does not become unambiguous during a backoff window. Every readonly call carrying a stale or guessed id paid **a second tmux round-trip and 100 ms of latency to fail identically** — and because `AuditMiddleware` sits outside the retry, the audit log recorded one call, so the doubled traffic was invisible.

This is **not new with #718.** `_resolve_pane` raises `PaneNotFound`, which is *already* a `LibTmuxException` — so panes have been over-retried all along. #718 merely widens it to every missing session and window, which is what made it impossible to keep ignoring. The fix closes the whole family, including the part that was already broken:

```
tests/test_middleware.py::test_readonly_retry_skips_deterministic_failures
  [TmuxObjectDoesNotExist]   FAILED on main   <- new with #718
  [MultipleObjectsReturned]  FAILED on main   <- new with #718
  [PaneNotFound]             FAILED on main   <- pre-existing
  [NoWindowsExist]           FAILED on main   <- pre-existing
  [BadSessionName]           FAILED on main   <- pre-existing
  [TmuxSessionExists]        FAILED on main   <- pre-existing
```

**2. An ambiguous target reached the agent with no way forward.** A window shared between sessions (a grouped session, or `link-window`) is listed once per session that holds it, so a name or index can match several rows and the lookup raises `MultipleObjectsReturned`. Today that surfaces as `Unexpected error: MultipleObjectsReturned:` with an empty message and a stack trace. Under #718 it would land somewhere worse: reclassified as *agent-correctable* (`expected: true`, WARNING, `"tmux error: …"`) **while still saying nothing about how to correct it**. It now names the fix — target the object by id.

## The fix

- `NON_RETRYABLE_EXCEPTIONS` in `middleware.py`, skipped in the retry decision. The check walks `__cause__` one hop, exactly as fastmcp's `_should_retry` does — necessary because `handle_tool_errors` re-raises every libtmux failure as an `ExpectedToolError` chained off the original, so at the middleware layer the real failure is only visible one hop down.
- An explicit `MultipleObjectsReturned` arm in `_map_exception_to_tool_error`, above the `LibTmuxException` arm, carrying a recovery suggestion.
- The not-found arm widened from `TmuxObjectDoesNotExist` to `ObjectDoesNotExist`, so a `QueryList` miss gets the same discovery hint as a tmux-id miss.

## Verification

Every new test **fails on `main` and passes here** — proven by reverting `src/` and re-running:

```
src/ = main   ->  8 failed
src/ = fixed  ->  all pass
```

`ruff format` · `pytest` · `ruff check --fix` · `mypy` (strict) · `just build-docs` — all green. **711 passed** (was 701).

The end-to-end test (`test_readonly_retry_skips_not_found_on_decorated_tool`) drives the real `list_sessions` tool through the production middleware stack, so it exercises the `handle_tool_errors` → `__cause__` → retry-decision chain rather than a unit stub. A unit test alone would pass even if the skip never fired in production.

## Before merge

1. libtmux#718 ships.
2. Drop `[tool.uv.sources]` from `pyproject.toml`.
3. Bump the floor: `libtmux>=<that release>,<1.0` (was `>=0.61.0`).
4. `uv lock`, re-run the gates.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,32 @@
 _Notes on upcoming releases will be added here_
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### Dependencies
+
+**Minimum `libtmux>=0.62.0`** (was `>=0.61.0`). Picks up libtmux 0.62.0, which
+re-parents {exc}`libtmux.exc.MultipleObjectsReturned` and
+{exc}`libtmux.exc.ObjectDoesNotExist` under {exc}`libtmux.exc.LibTmuxException`.
+The retry and error-mapping fixes below rely on that hierarchy. (#98)
+
+### Fixes
+
+**A stale id fails once instead of twice**
+
+Readonly tools retried on any {exc}`libtmux.exc.LibTmuxException`, the right
+trigger for the transient socket errors libtmux wraps but wrong for the rest of
+that family — a pane that does not exist will not exist on the second look, and
+an ambiguous match does not become unambiguous during a backoff window. A lookup
+that a retry cannot change now fails on the first attempt, without the redundant
+tmux round-trip and its backoff delay. (#98)
+
+**An ambiguous target says how to disambiguate it**
+
+A window shared between sessions — through a grouped session or `link-window` —
+is listed once per session that holds it, so a name or index can match more than
+one row and the lookup raises {exc}`libtmux.exc.MultipleObjectsReturned`. That
+reached the agent with no recovery hint. It now says to target the object by id
+instead. (#98)
+
 ## libtmux-mcp 0.1.0a17 (2026-07-11)
 
 libtmux-mcp 0.1.0a17 makes best-effort shell-history suppression the default for {tooliconl}`run-command` — calls that omit `suppress_history` now inherit the server default, and setting {envvar}`LIBTMUX_SUPPRESS_HISTORY` to `0` restores the previous behavior. Spawned shells gain the same hygiene: {tooliconl}`create-session`, {tooliconl}`create-window`, {tooliconl}`split-window`, and {tooliconl}`respawn-pane` accept `suppress_persistent_history=true` for best-effort no-disk history controls. {tooliconl}`create-window` and {tooliconl}`split-window` also accept per-process `environment` mappings, and the MCP install picker now covers Grok CLI and Antigravity.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ include = [
 ]
 
 dependencies = [
-  "libtmux>=0.61.0,<1.0",
+  "libtmux>=0.62.0,<1.0",
   "fastmcp>=3.4.2,<4.0.0",
 ]
 
--- a/src/libtmux_mcp/_utils.py
+++ b/src/libtmux_mcp/_utils.py
@@ -1001,11 +1001,20 @@ def _map_exception_to_tool_error(fn_name: str, e: BaseException) -> ToolError:
         return ExpectedToolError(str(e))
     if isinstance(e, exc.BadSessionName):
         return ExpectedToolError(str(e))
-    if isinstance(e, exc.TmuxObjectDoesNotExist):
+    if isinstance(e, exc.ObjectDoesNotExist):
         return ExpectedToolError(
             f"Object not found: {e}",
             suggestion=(
                 "Call list_sessions / list_windows / list_panes to discover valid ids."
+            ),
+        )
+    if isinstance(e, exc.MultipleObjectsReturned):
+        return ExpectedToolError(
+            f"Ambiguous target: {e}",
+            suggestion=(
+                "A window shared between sessions is listed once per session that "
+                "holds it, so a name or index can match more than one row. Target "
+                "it by id (session_id / window_id / pane_id) instead."
             ),
         )
     if isinstance(e, exc.PaneNotFound):
--- a/src/libtmux_mcp/middleware.py
+++ b/src/libtmux_mcp/middleware.py
@@ -666,6 +666,45 @@ class AuditMiddleware(Middleware):
 DEFAULT_RESPONSE_LIMIT_BYTES = 1_000_000
 
 
+#: Failures a retry cannot fix. Each one names a thing that is not there, or a
+#: request that cannot succeed as written, so re-running it buys a second tmux
+#: round-trip and a backoff window in order to fail identically. They all
+#: descend from :exc:`libtmux.exc.LibTmuxException`, which is the retry
+#: trigger, so without this set they would all be retried.
+#:
+#: Order the entries most-general-first when reading: ``ObjectDoesNotExist``
+#: already covers :exc:`libtmux.exc.TmuxObjectDoesNotExist`.
+NON_RETRYABLE_EXCEPTIONS: tuple[type[Exception], ...] = (
+    libtmux_exc.ObjectDoesNotExist,
+    libtmux_exc.MultipleObjectsReturned,
+    libtmux_exc.PaneNotFound,
+    libtmux_exc.NoWindowsExist,
+    libtmux_exc.BadSessionName,
+    libtmux_exc.TmuxSessionExists,
+    libtmux_exc.TmuxCommandNotFound,
+)
+
+
+class _SkipDeterministicFailures(RetryMiddleware):
+    """A :class:`RetryMiddleware` that declines to retry what cannot succeed.
+
+    fastmcp decides whether to retry in ``_should_retry``, matching the error
+    and, one hop down, its ``__cause__`` -- which is where the real failure
+    lives, because ``handle_tool_errors`` re-raises every libtmux error as an
+    ``ExpectedToolError`` chained off the original. This narrows that decision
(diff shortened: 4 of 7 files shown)


tony:
### Code review

Found 6 issues. The first two are blockers.

1. **The module does not load on the floor it declares.** `libtmux.exc.MultipleObjectsReturned` does not exist on released libtmux 0.61.0 — `exc.py` there re-exports only `ObjectDoesNotExist`. The floor is `libtmux>=0.61.0,<1.0`, which permits it. `NON_RETRYABLE_EXCEPTIONS` is built at module level, `server.py` imports `middleware`, and `main()` imports `server` — so the console script dies at startup with `AttributeError`, which `main()`'s `except ImportError:` guard does not catch. The git pin masks this: the `winlinks` branch still self-reports `__version__ = "0.61.0"`, so nothing — uv, mypy, or CI — will ever warn. The floor bump must land in the **same commit** that drops `[tool.uv.sources]`, not merely "before merge".

https://github.com/tmux-python/libtmux-mcp/blob/a07a6e16468268764620a7d0c01cb44a49084862/src/libtmux_mcp/middleware.py#L676-L686

2. **Second, independent break — inside the error handler.** `_utils.py` imports fine on 0.61.0 (the reference is in a function body), so fixing `middleware.py` alone would not help. This arm sits above `PaneNotFound` and the `LibTmuxException` catch-all, so on released libtmux **every** exception reaching the mapper raises `AttributeError` from inside the handler, destroying the real diagnostic instead of reporting it.

https://github.com/tmux-python/libtmux-mcp/blob/a07a6e16468268764620a7d0c01cb44a49084862/src/libtmux_mcp/_utils.py#L1010-L1020

3. **The recovery suggestion is backwards — it tells the agent to do the thing that raises.** For the scenario it names (a window shared between sessions), `window_id` and `pane_id` are the duplication *victims*, not the disambiguator: the duplication key is `(window_id, session_id)`. `_resolve_window` does `server.windows.get(window_id=...)` over `list-windows -a`, which returns one row per holding session. Probed live on a `link-window`ed window:

```
server.windows.get(window_id='@0')   -> RAISES  Multiple objects returned (2): window_id='@0'
server.panes.get(pane_id='%0')       -> RAISES  Multiple objects returned (2): pane_id='%0'
session.windows.get(window_id='@0')  -> OK
```

An agent that follows the suggestion re-issues by id, gets the identical error, and loops — which is the opposite of the "agent-correctable" contract this function's own docstring promises.

https://github.com/tmux-python/libtmux-mcp/blob/a07a6e16468268764620a7d0c01cb44a49084862/src/libtmux_mcp/_utils.py#L1010-L1020

4. **`Fixes #97` will auto-close an issue this PR does not fix.** GitHub ignores the parenthetical qualifier. [#97](https://github.com/tmux-python/libtmux-mcp/issues/97) asks for the resolvers to route through `Pane.from_pane_id` / `Window.from_window_id`; this PR implements only what #97 itself calls the "interim mitigation". `_resolve_window`/`_resolve_pane` still filter a server-wide listing and still raise on a grouped session, including the self-kill guards #97 names. Should be `Refs #97`.

5. **`NON_RETRYABLE_EXCEPTIONS` is a fail-open denylist, and is already incomplete.** The trigger stays the whole `LibTmuxException` family and this subtracts seven classes, so anything libtmux adds later defaults to retryable. It already misses `NotInsideTmux` — **new on the very branch this PR exists to validate** — and the `OptionError` family (`InvalidOption`, `AmbiguousOption`, `UnknownOption`), which is raised by the readonly `show_option` tool today. A mistyped option name is exactly the deterministic failure this PR is about, and it still pays a retry. The original design commit ([`707eed8`](https://github.com/tmux-python/libtmux-mcp/commit/707eed8)) states the intent as an allowlist: "libtmux wraps the subprocess failures we actually want to retry (socket EAGAIN, transient connect errors)". Inverting to an allowlist is fail-safe and closer to that.

6. **CHANGES uses libtmux's changelog convention, not this repo's** (AGENTS.md says "every distinct deliverable gets a `**Bold subheading**`" and "PR refs `(#NN)` sit at the end of each deliverable's prose paragraph"). These two `####` headings are the only ones in the file; every prior deliverable back to 0.1.0a10 uses `**Bold**` with the ref at paragraph end. The mandated `### Dependencies` section recording the required libtmux floor is also missing.

https://github.com/tmux-python/libtmux-mcp/blob/a07a6e16468268764620a7d0c01cb44a49084862/CHANGES#L10-L12

One note worth weighing rather than fixing: this PR subclasses and overrides fastmcp's private `RetryMiddleware._should_retry`. [`8b102bb`](https://github.com/tmux-python/libtmux-mcp/commit/8b102bb) exists because fastmcp changed that method's semantics **in a patch release**, and this project shipped a silent production no-op with a green suite as a result. The `<4.0.0` ceiling gives a leading-underscore method no protection. The new end-to-end test would catch a dead override on a dep bump — a real improvement over the state that produced `8b102bb` — but inlining the ~25-line retry loop would remove the coupling entirely.

Issue #54, opened 2026-05-16
wait_for_text: expose risk_band_warned: bool in WaitForTextResult for clients that don't subscribe to log notifications


tony:
**Type:** enhancement · **Tier:** deferred · **Tool:** `wait_for_text`

## Problem

When `wait_for_text` polls a pane near `history-limit`, it emits a one-shot `ctx.warning()` notification: "wait_for_text correctness is best-effort here. For deterministic synchronization use wait_for_channel." That warning routes through MCP client notifications. Some clients, stdio bridges, and programmatic callers do not surface those notifications to the agent or caller that receives the final tool result.

The current `WaitForTextResult` reports `found`, `matched_lines`, `pane_id`, and `elapsed_seconds`, but it does not include a structured indication that polling entered the history-limit risk band. The caller can therefore receive `found=True` or `found=False` without knowing that the answer was best-effort.

## Proposed fix

Add a structured field to `WaitForTextResult`:

```python
class WaitForTextResult(BaseModel):
    found: bool = ...
    matched_lines: list[str] = ...
    pane_id: str = ...
    elapsed_seconds: float = ...
    risk_band_warned: bool = Field(
        default=False,
        description=(
            "True if polling entered the history-limit trim-risk band "
            "at any point during the wait. Indicates wait_for_text's "
            "match result is best-effort; compose tmux wait-for -S "
            "with wait_for_channel for deterministic authored-command "
            "completion."
        ),
    )
```

The flag is already tracked inside `wait_for_text` as `warned_risk_band`, which gates the one-shot warning. Surface that value in the returned model.

## Trade-offs

- This is an additive public response-model field. The project is still `0.1.x`, but once shipped it should be treated as part of the tool contract.
- Tests should extend `test_wait_for_text_warns_in_history_limit_risk_band` and `test_wait_for_text_warns_when_already_in_risk_band` to assert `result.risk_band_warned is True`, and representative non-risk-band tests should assert the default is `False`.
- Docs and CHANGES should mention that the structured field mirrors the notification for clients that do not subscribe to or expose log messages.

## Why this remains separate

- #50 covers a state/capture race inside `wait_for_text`.
- #51 covers opt-in same-row matching.
- #52 covers polling pressure and adaptive backoff.
- #53 covers `wait_for_content_change` lifecycle parity.
- #60 proposes a new non-blocking observation primitive.
- #61 is the umbrella architecture issue.

[...]

PR #100, opened 2026-07-22
Fix wait_for_text's blind spot and bound every wait

tony:
## Summary

- **Fix a blind spot that made the tool useless for its own headline case.** `wait_for_text` anchored one row *below* where the cursor sat when the call began — and on a quiescent pane that is exactly where the next line of output lands, because the cursor sits at the end of the prompt. A daemon printing a single `ready` line could not be matched at all: the wait burned its whole budget and then reported that the pane had been silent. The anchor now lands on that row, and the row's pre-existing content is suppressed by value instead of by index, so carriage-return spinners and single-line status updates become matchable too.
- **Bound every wait, and let the caller see the bound.** `LIBTMUX_MCP_WAIT_MAX_SECONDS` (30 s default, clamped to `[1, 120]`) caps `wait_for_text`, `wait_for_channel` and `run_command`. An over-large `timeout` is clamped rather than rejected, and the result reports what was actually enforced — the agent learns the policy from the result instead of from a failed call.
- **Replace `pattern: str` with `patterns: list[str] | None`, and add `stop`.** `patterns=null` waits for any new output, subsuming the removed `wait_for_content_change`. A `stop` list ends the wait in milliseconds on a known failure marker instead of burning the budget.
- **Make the result explain itself.** One `outcome` enum replaces a scatter of booleans, alongside which entry fired, whether anything new was written, whether the match was text already on screen at entry, and a byte-bounded tail. A timeout now distinguishes "the command never ran" from "output arrived and did not match" from "a pager owned the pane".
- **Stop waits hanging, leaking, or reporting a result they did not get.** A wedged tmux could hang the process and Ctrl-C indefinitely; a cancelled call left its tmux child running for the rest of its timeout; a retried or batched wait could exceed its own ceiling; and `wait_for_channel` reported success when the tmux server shut down without ever signalling, because `tmux wait-for` exits 0 for both.
- **Make failures name what failed.** A missing dependency, an unresolvable pane target, and the client behind a bad argument were all reported without naming the thing at fault.

### `src/libtmux_mcp/tools/pane_tools/wait.py`

The core. Entry-row anchoring with content-based suppression of pre-existing text, `patterns`/`stop` handling, the alternate-screen guard, and every tmux call spawned as a deadline-bounded async subprocess rather than reaching tmux through a worker thread.

### `src/libtmux_mcp/_wait_policy.py`, `server.py`

The ceiling policy, resolved from the environment and injected at tool-registration time, mirroring the existing history-default pattern. `server.py` also fails at startup rather than silently dropping the `is_caller` agent-context instructions when the MCP instruction budget is tight, while still degrading gracefully for pathological runtime values.

### `_utils.py`, `middleware.py`, `batch_tools.py`, `wait_for_tools.py`

The `self-bounded` tag and its enforcement: excluded from readonly retry, rejected per-operation by every batch wrapper, and applied to `wait_for_channel` and `run_command` as well as `wait_for_text`. `wait_for_channel` gains the liveness re-probe that distinguishes a signal from a server that went away.

### `models.py`, `state.py`

`WaitForTextResult` reshaped around the `outcome` enum; `ContentChangeResult` removed. `_PaneState` gains `alternate_on`, tolerant of older tmux that omits the field.

### `prompts/recipes.py`, docs

The `interrupt_gracefully` recipe no longer treats a `^C` echo as evidence that a program survived SIGINT — the terminal echoes it on delivery either way, so the marker fired on the success path and the recipe's next step is escalation. It now brackets the interrupt with `get_pane_info` and trusts `pane_current_command`. A new **Waiting** topic covers which wait to reach for, how the "is this new" predicate works, and what a wrong pattern costs.

## Design decisions

- **Thread-free, not just timeout-bounded.** A private `ThreadPoolExecutor` with `shutdown(wait=False)` was considered and rejected: `concurrent.futures.thread._python_exit` joins every pool worker untimed regardless, so no thread-based arrangement survives a wedged tmux at interpreter exit. Only owning a killable subprocess does.
- **Kill-then-cancel-then-bounded-reap.** The textbook `wait_for(communicate())` then `kill()` then `await wait()` deadlocks when a wedged tmux's grandchild still holds the stdout pipe. The reader task is torn down first and the final reap is time-boxed; the loop's child watcher reaps the pid regardless.
- **Suppress the entry row by content, not by index.** Skipping the row by index is what created the blind spot. Filtering it by the text that was on it at entry keeps stale paint from matching while letting genuinely new text on the same row through. Verified in both directions: a broad pattern still does not match the shell prompt that was already there.
- **A stale match does not end the wait.** `matched_at_entry` is known before the poll loop starts, so returning immediately looks like free latency. It would break the commonest agent loop — run, wait for marker, re-run, wait for the same marker — so the wait holds out for a fresh occurrence and says so in the result.
- **The ceiling bounds the agent's turn, not the transport.** The original rationale was that a long wait stalls the shared MCP connection. It does not: the wait tools await throughout and FastMCP serves the connection concurrently — measured, a tool awaiting 6 s served 58 interleaved calls at a 3.4 ms median, against exactly one at 6014 ms for a control that blocked the event loop. The ceiling is unchanged and still correct, but for the honest reason: MCP gives an agent no way to abandon a call mid-flight.
- **`outcome` as one enum, not three booleans.** The output schema is re-sent on every request, so a field no agent branches on is a permanent token tax.
- **Background tasks were evaluated and rejected.** Both the SEP-1686 shape and the SEP-2663 `fastmcp-tasks` alpha were trialled against this tool. Neither is reachable — no shipping client negotiates the capability — and both drop the numeric progress this tool emits. Recorded in the Waiting topic so it is not re-litigated from intuition.

## Verification

No caller pattern reaches a tmux format string — tmux treats `#` and `}` structurally, so an ordinary regex quantifier would corrupt field parsing:

```
rg -n 'C/|display-message.*\{pattern' src/libtmux_mcp/tools/pane_tools/wait.py
```

Nothing on the wait path reaches tmux through a thread:

```
rg -n 'to_thread\(|subprocess\.run\(' src/libtmux_mcp/tools/pane_tools/wait.py
```

The removed tool and its result model are gone as symbols:

```
rg -n '^\s*(def|class)\s+(wait_for_content_change|ContentChangeResult)\b' src/ docs -g '!_build'
```

## Test plan

- [x] `uv run ruff format --check .`, `uv run ruff check .`, `uv run mypy` clean
- [x] `uv run pytest --reruns 0` green with flake-masking reruns disabled
- [x] `just build-docs` succeeds; removing the `wait-for-content-change` page leaves no broken refs
- [x] CI green across the full tmux matrix — 3.2a, 3.3a, 3.4, 3.5, 3.6 and master — so the grid arithmetic is not validated against one tmux only
- [x] `test_wait_for_text_sees_the_entry_cursor_row` — a bare line, an unterminated prompt, a trailing-line case and a carriage-return rewrite all match on the entry row; two controls pin the other direction
- [x] `test_wait_for_text_does_not_match_the_prompt_on_the_entry_row` — the falsifier: a broad pattern must not self-match the prompt that was already on the row
- [x] `test_wait_for_text_waits_for_a_fresh_occurrence` — a stale match on screen does not short-circuit the wait
- [x] `test_wait_path_uses_no_worker_threads` — the wait path spawns no threads
- [x] `test_wait_tools_do_not_block_event_loop` — the loop survives a genuinely wedged tmux; mutation-tested to fail if a blocking spawn returns
- [x] `test_wait_for_text_bounds_every_tmux_call` — every spawned tmux call is deadline-bounded
- [x] `test_wait_for_channel_detects_a_vanished_server` — a clean `kill-server` and a SIGTERM are both reported as errors, with a live-server control that must still succeed
- [x] `test_wait_resolver_matches_the_canonical_resolver` — the wait's private resolver agrees with `_utils._resolve_pane` on precedence and on the exception a miss raises
- [x] A wedged tmux lets the process exit promptly, and under SIGINT, where the prior code hung indefinitely
- [x] Native pane resolution verified equivalent to libtmux across all `wait_for_text` targeting cases, including the linked-window `MultipleObjectsReturned` trap
- [x] Verified end to end over a real MCP stdio session and through a real CLI agent, with the mutation asserted independently against the target tmux socket rather than trusting the agent's report


tony:
### Code review

Found 6 issues (1 confirmed by repro):

1. Cancelling a wait orphans the in-flight tmux child. The cancellation lands at `await asyncio.wait({task}, ...)`, which does not cancel `task`; the `except asyncio.CancelledError` only guards the synchronous `task.result()` below it, so `_kill_and_reap` never runs. I reproduced it against a wedged tmux: `CancelledError propagated in 0.00s` but `ORPHANED fake-tmux alive: 1`, and the leaked child re-introduces the interpreter-shutdown hang this PR set out to fix. The window is largest under a wedged tmux — exactly the target scenario. Wrap the `await asyncio.wait(...)` in `try/except CancelledError` (or `try/finally`) that reaps. (bug)

https://github.com/tmux-python/libtmux-mcp/blob/6bd82ce873e454038b066273e7fd5e175000f75a/src/libtmux_mcp/tools/pane_tools/wait.py#L238-L252

2. `interrupt_gracefully` passes `stop=["^C", ...]` with `regex=True`. As a regex, `^C` is start-anchored `C`: `re.search("^C", "^C")` is `False` (it never matches the literal `^C` echo it targets) and `re.search("^C", "Compiling")` is `True` (it false-fires `outcome="stopped"` on any line starting with `C`). Escape it (`\^C`) or drop `regex=True` for that marker. (bug)

https://github.com/tmux-python/libtmux-mcp/blob/6bd82ce873e454038b066273e7fd5e175000f75a/src/libtmux_mcp/prompts/recipes.py#L166-L171

3. `build_dev_workspace` step 5 tells the agent `wait_for_text(patterns=null)` confirms "a vim splash screen," but this PR's alternate-screen guard suppresses matching while `alternate_on` and returns `outcome="alternate_screen"`, `found=False`. vim uses the alternate screen, so the stated guarantee is now false. (bug)

https://github.com/tmux-python/libtmux-mcp/blob/6bd82ce873e454038b066273e7fd5e175000f75a/src/libtmux_mcp/prompts/recipes.py#L138-L143

4. The `wait_for_text` docstring tells the agent the result reports `suppressed_stale_match`, but no such field exists — the shipped field is `matched_at_entry`. (stale doc on a changed line)

https://github.com/tmux-python/libtmux-mcp/blob/6bd82ce873e454038b066273e7fd5e175000f75a/src/libtmux_mcp/tools/pane_tools/wait.py#L517-L520

5. The same docstring states "Alternate screen / pagers are not handled ... this tool does not detect it," which contradicts the alternate-screen handling this PR adds. (stale doc on a changed line)

https://github.com/tmux-python/libtmux-mcp/blob/6bd82ce873e454038b066273e7fd5e175000f75a/src/libtmux_mcp/tools/pane_tools/wait.py#L588-L592

6. The poll-loop comment still says the tmux reads are "blocking subprocess calls" pushed to "the default executor," but this PR makes the path a native async subprocess with no executor and no thread — the comment describes the code it replaced. (stale comment on a changed line)

https://github.com/tmux-python/libtmux-mcp/blob/6bd82ce873e454038b066273e7fd5e175000f75a/src/libtmux_mcp/tools/pane_tools/wait.py#L712-L716

tony:
### Code review

Found 3 issues (re-checked against the current, rebased branch):

1. Cancelling a wait can orphan the in-flight tmux child. The `except asyncio.CancelledError` only wraps `task.result()`, not the `await asyncio.wait({task}, timeout=budget)` above it. When the MCP client hangs up mid-wait, cancellation almost always lands on that `asyncio.wait` call, so `_kill_and_reap` is skipped and the tmux subprocess (and its `communicate()` future) leaks -- the exact teardown this PR set out to guarantee.

https://github.com/tmux-python/libtmux-mcp/blob/29934119541df083a2ba99ce7197826dfbb380ff/src/libtmux_mcp/tools/pane_tools/wait.py#L238-L254

2. The `wait_for_text` docstring says alternate-screen panes are not detected, but this PR detects them. The Notes section states "this tool does not detect it," while the PR adds `state.alternate_on` handling that returns `outcome="alternate_screen"`. The agent-facing docs now understate the tool's behavior.

https://github.com/tmux-python/libtmux-mcp/blob/29934119541df083a2ba99ce7197826dfbb380ff/src/libtmux_mcp/tools/pane_tools/wait.py#L589-L594

3. The docstring references a field that no longer exists. It reports a stale entry match via `suppressed_stale_match`, but this PR renamed that field to `matched_at_entry` in `models.py`, and `suppressed_stale_match` appears nowhere else in the codebase.

https://github.com/tmux-python/libtmux-mcp/blob/29934119541df083a2ba99ce7197826dfbb380ff/src/libtmux_mcp/tools/pane_tools/wait.py#L517-L520

Generated with [Claude Code](https://claude.ai/code)

tony:
### Code review

Found 5 issues:

1. `run_command` honours the wait ceiling but is not tagged `TAG_SELF_BOUNDED`, so batches can still multiply it. This PR adds `effective_timeout = min(timeout, _wait_ceiling_seconds())` to `run_command`, putting it in the same class as `wait_for_text` and `wait_for_channel` — both of which were given `TAG_SELF_BOUNDED` so `batch_tools._get_allowed_tool_tier` rejects them per-operation. `run_command` is still registered with `tags={TAG_MUTATING}` alone, so `call_mutating_tools_batch` / `call_destructive_tools_batch` accept it, and `MAX_BATCH_OPERATIONS` of them run serially with no aggregate deadline — up to 1000x the ceiling. That is exactly the multiplication the `wait_for_tools.py` rationale describes.

https://github.com/tmux-python/libtmux-mcp/blob/3a95d898b84ac997fd5416c7012137b8c0133177/src/libtmux_mcp/tools/pane_tools/__init__.py#L93-L97

2. Still open from the first review round: `interrupt_gracefully` passes `stop=["^C", "Interrupt"]` with `regex=True`. As a regex, `^C` is start-anchored `C` — `re.search("^C", "^C")` is `False`, so it never matches the literal `^C` echo it targets, and `re.search("^C", "Compiling")` is `True`, so it false-fires `outcome="stopped"` on any line starting with `C`. Escape it (`\^C`) or drop `regex=True` for that marker.

https://github.com/tmux-python/libtmux-mcp/blob/3a95d898b84ac997fd5416c7012137b8c0133177/src/libtmux_mcp/prompts/recipes.py#L167-L172

3. `WaitForTextResult.risk_band_warned` is removed but not documented under `### Breaking changes` (AGENTS.md says "Buried breaking changes — they get their own subheading at the top of the entry"). The field was added for #54 precisely because "clients that do not surface MCP warning notifications can still detect that matching was best-effort" — `wait.py` still computes `warned_risk_band` but no longer returns it, so those clients lose that signal silently. The unreleased breaking-changes section covers only the `patterns` rename and the `wait_for_content_change` removal.

https://github.com/tmux-python/libtmux-mcp/blob/3a95d898b84ac997fd5416c7012137b8c0133177/CHANGES#L8-L24

4. The `wait_for_channel` comment says it mirrors `wait_for_text`, which this PR made false. Commit f62de4e rewrote the `wait_for_text` path to native `asyncio` subprocesses specifically because a thread blocked in libtmux's untimed `Popen.communicate()` cannot be cancelled and hangs interpreter shutdown. `wait_for_channel` still uses `asyncio.to_thread(subprocess.run, ...)` — the pattern that commit removed — directly under a comment claiming it mirrors the rewritten one.

https://github.com/tmux-python/libtmux-mcp/blob/3a95d898b84ac997fd5416c7012137b8c0133177/src/libtmux_mcp/tools/wait_for_tools.py#L173-L182

5. `RunCommandResult.effective_timeout` is cited with `{class}` instead of `{attr}` in CHANGES (AGENTS.md says "attributes use `{attr}`"). Precedent in the same file: `` {attr}`~libtmux_mcp.models.SessionInfo.active_pane_id` ``.

https://github.com/tmux-python/libtmux-mcp/blob/3a95d898b84ac997fd5416c7012137b8c0133177/CHANGES#L76-L78

Generated with [Claude Code](https://claude.ai/code)

tony:
### Code review

Fresh pass over the six commits since the last review (`3a95d89..dd33e4d`). The five findings from that round are all genuinely fixed — I re-verified each, including that the emitted recipe marker is `\^C` (single backslash, matches literal `^C`, not `Compiling`) and that `run_command` is now rejected by all three batch wrappers.

Found 2 issues:

1. The new breaking-change entry ships migration advice that does not detect what it claims to. It offers `outcome == "timeout" and not matched_at_entry` as the "closest substitute" for the removed `risk_band_warned`, in a copyable `# Before` / `# After` block. But `warned_risk_band` is set from `state.history_size >= risk_floor` on any poll tick, independent of how the wait ends — so a wait that *matches* while polling through the trim-risk band is never flagged (the check requires `outcome == "timeout"`), an ordinary timeout on a pane nowhere near its history limit is flagged falsely, and panes with `history_limit == 0` skip the risk-band check entirely while still satisfying the substitute. Suggest saying plainly that there is no result-field equivalent and pointing at the warning notification and `wait-for-channel`, rather than offering a check that does not correlate.

https://github.com/tmux-python/libtmux-mcp/blob/dd33e4d3039ba7df8813ea0bef22010717d4953f/CHANGES#L45-L62

2. Cancelling a wait still orphans the in-flight tmux child. Third round flagging this one, so noting it is still verbatim unchanged at head rather than re-arguing it: `await asyncio.wait({task}, timeout=budget)` on line 239 sits outside the `try:`, which opens on line 247 and guards only `task.result()`. A client hanging up mid-wait lands its cancellation on the `asyncio.wait` far more often than on the synchronous `task.result()`, so `_kill_and_reap` is skipped and the tmux subprocess leaks — the interpreter-shutdown hang this PR set out to remove. Wrapping the `await asyncio.wait(...)` in the same `try` (or a `try/finally` that reaps) closes it.

https://github.com/tmux-python/libtmux-mcp/blob/dd33e4d3039ba7df8813ea0bef22010717d4953f/src/libtmux_mcp/tools/pane_tools/wait.py#L237-L254

Below the reporting threshold but still open, both in `wait.py` and both flagged in earlier rounds: the docstring Notes section still says alternate screen / pagers "are not handled" and "this tool does not detect it", and the poll-loop comment still describes pushing blocking reads to "the default executor" — three lines below a docstring stating nothing runs on a worker thread.

Generated with [Claude Code](https://claude.ai/code)

PR #103, opened 2026-07-24
Add MCP CLI-testing skill and mcp_swap doctor/--env

tony:
## Summary

- **Add** a `testing-mcp-with-cli-agents` skill — a verified method for driving all six installed agent CLIs (Claude, Codex, Cursor, Gemini, Grok, agy) against the server on isolated tmux sockets and asserting ground truth, instead of trusting unit tests alone.
- **Add** a read-only `mcp_swap.py doctor` subcommand that reports the effective swap environment: which server name each CLI points at (and a mismatch when the repo is registered under a name other than the derived default), un-reverted swaps and orphaned backups on disk, a state entry whose backup has gone missing (so `revert` would fail), and auth-overriding env vars such as `OPENAI_API_KEY`.
- **Add** `use-local --env KEY=VALUE` (repeatable) to write env — e.g. an isolated `LIBTMUX_SOCKET` — into the server entry at swap time, layered over preserved env.
- **Add** a naming hint on `use-local` when the repo is registered under a server name other than the derived default, so a bare run doesn't silently no-op on a missing entry.

### Dev tooling — `scripts/mcp_swap.py`

- **`doctor`**: cross-CLI entry inventory, server-name mismatch hint, outstanding-swap and orphaned-backup health, missing-backup detection, and auth-env override warnings. Read-only; deliberately does not model each CLI's config-merge behaviour (that is version-specific and lives in the skill docs).
- **`use-local --env`**: repeatable `KEY=VALUE`, merged over any preserved env, explicit wins.
- **naming hint**: surfaced on `use-local` and in `doctor` via a reusable scan for repo-pointing entries registered under another name.

### Skill — `.agents/skills/testing-mcp-with-cli-agents/`

- `SKILL.md`: the three-socket isolation model, three fidelity layers (direct smoke, headless one-shot, interactive send-keys), the cancellation/orphan test, and trunk-vs-branch comparison.
- `references/cli-matrix.md`: per-CLI isolation lever, cheapest discovery proof, headless invocation, and approval-bypass flag — verified across all six CLIs.
- Linked into `.claude/skills/` via a committed relative symlink; the source lives in the tool-neutral `.agents/` tree (`.claude/` is git-ignored).

## Design decisions

- **`doctor` reports observable state, not CLI internals**: it inventories config entries, swap state, backups, and env vars — all stable and observable — but does not encode version-specific config-merge behaviour, which would rot. That knowledge lives in the skill's cli-matrix with a dated caveat.
- **Server-agnostic implementation**: the new helpers derive the server slug from `pyproject.toml` and hardcode no name, so the same logic ports to sibling repos that vendor `mcp_swap.py`.
- **Zero-mutation isolation preferred for tests**: the skill steers testers toward each CLI's throwaway config-home / project-config lever over `mcp_swap use-local`, since a swap mutates real configs; `doctor` and `--env` make the swap path safer when it is used.

## Test plan

- [x] `uv run ruff format . && uv run ruff check .`
- [x] `uv run mypy src tests`
- [x] `uv run pytest tests/test_mcp_swap.py` — env injection, naming hint, and `doctor` (mismatch, missing-backup, orphaned-backup) coverage
- [x] `uv run pytest` — full suite green
- [x] `just build-docs` — CHANGES renders
- [x] `uv run scripts/mcp_swap.py doctor --server tmux` — read-only smoke against the real environment

### DIFF
--- a/.agents/skills/testing-mcp-with-cli-agents/SKILL.md
+++ b/.agents/skills/testing-mcp-with-cli-agents/SKILL.md
@@ -1,0 +1,183 @@
+---
+name: testing-mcp-with-cli-agents
+description: >-
+  Test an MCP server by driving real CLI agents (Claude, Codex, Cursor, Gemini,
+  Grok, agy) against it, using isolated tmux sockets and send-keys instead of
+  trusting unit tests alone. Use this whenever verifying MCP-server behavior
+  end-to-end, checking that a local branch or checkout works across installed
+  agent CLIs, comparing trunk-vs-branch MCP behavior, driving an interactive
+  agent TUI to exercise approval flows or cancellation, reproducing a tmux-MCP
+  bug through a live client, or wiring a checkout into the CLIs with mcp_swap.
+  Reach for it even when the user only says "test the MCP", "does the branch
+  work in the agents", "drive the CLI to call the tool", or "check it across
+  Codex/Gemini/Cursor" without naming tmux or sockets explicitly.
+---
+
+# Testing an MCP server through real CLI agents
+
+Unit tests prove the server's internals; they don't prove a real agent can
+discover a tool, get past its approval gate, call it, and survive cancelling it
+mid-flight. This skill exercises that whole path by pointing installed CLI
+agents at a checkout and driving them — with the tmux tool surface (libtmux-mcp)
+as the running example, though the shape generalizes to any MCP server.
+
+## The core idea: three tmux servers, never one
+
+The single biggest mistake is running everything on one tmux socket. Keep three
+distinct servers, each on its own socket, and the whole thing becomes safe and
+observable:
+
+| Role | Socket | Who touches it | Why separate |
+|---|---|---|---|
+| Your real session | default | you, interactively | must never be mutated by a test |
+| Harness | `tmux -L cli-harness` | the driver: `send-keys` prompts in, `capture-pane` render out | isolates the CLI TUI you're driving |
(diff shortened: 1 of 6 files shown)


tony:
### Code review

Found 1 issue:

1. `use-local --env` (and a changed `--entry`) is silently dropped when the entry already points at this repo. The "already local (this repo)" short-circuit `continue`s before the env merge just below it, so on a machine whose configs already point local — which `doctor` reports is the case for all six CLIs here — `use-local --server tmux --env LIBTMUX_SOCKET=…` prints "already local — no change" and never writes the env. The isolation the skill documents (`--env LIBTMUX_SOCKET`) is then silently not applied. The merge's own comment — "without this merge a swap would silently drop client-side settings (LIBTMUX_SOCKET, ...)" — describes exactly the case the guard above it bypasses. Reproduced against a sandbox config already pointing local: the flag is a confirmed no-op.

https://github.com/tmux-python/libtmux-mcp/blob/c8fdc4209e0cf67875c2c967289aa7a9ea193213/scripts/mcp_swap.py#L871-L889

Generated with [Claude Code](https://claude.ai/code)

tony:
Follow-up: the `use-local --env` finding (A1) from the earlier review is resolved in the current head.

The already-local short-circuit in `cmd_use_local` now only fires when the requested `--env` is already satisfied, so `--env` (e.g. an isolated socket/data-dir) is written even when the config already points at this repo. A regression test covers both the write path and the idempotent no-op, and it was verified end-to-end against an isolated tmux socket (booted server, ground-truth confirmed, real config untouched). Two smaller items from that pass — a non-dict config entry no longer aborts `doctor`/`use-local`, and the naming hint no longer false-fires when the repo is registered under both the derived and another name — landed in the same commit.

PR #110, opened 2026-07-25
mcp(fix[wait]): Stop using the deprecated MCP Logging capability

### LINKED ISSUE #109: Migrate wait_for_text off the deprecated MCP Logging capability (SEP-2577)
## Summary

MCP has deprecated the protocol-level **Logging** capability. `wait_for_text` is the only tool in this server that uses it, via three `_maybe_log` calls that emit `notifications/message`. This server is stdio-only, so the specified migration path — log to `stderr` — is a direct swap onto the stdlib loggers the codebase already uses everywhere else.

## Upstream status

MCP adopted a [feature lifecycle and deprecation policy](https://modelcontextprotocol.io/community/feature-lifecycle) in [SEP-2596](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2596), which defines Active/Deprecated/Removed states, a twelve-month minimum deprecation window, and a canonical [deprecated features registry](https://modelcontextprotocol.io/specification/draft/deprecated).

[SEP-2577](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577) (merged 2026-05-15) deprecates Roots, Sampling, and **Logging**, effective with the `2026-07-28` revision. The registry gives Logging's migration path as *"Log to `stderr` for stdio transports; use OpenTelemetry for observability"*, and the rationale as *"overlaps with stderr and OpenTelemetry"*. Earliest removal is the first revision released on or after 2027-07-28.

Separately, [SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575) removes `logging/setLevel` in the same revision and moves log level to a per-request `_meta` key, with a hard constraint: *"servers MUST NOT emit `notifications/message` for requests that did not include this field"*. Under `2026-07-28` the current unconditional emissions are therefore not merely deprecated but non-conforming.

This is not urgent — the pinned SDK negotiates `2025-11-25` (`mcp.types.LATEST_PROTOCOL_VERSION`), where Logging is still Active — but the migration is small and there is no reason to carry it.

## Affected code

All three call sites are in `wait_for_text`:

| Site | What it reports | Already reaches the agent? |
| --- | --- | --- |
| [`wait.py#L434`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/tools/pane_tools/wait.py#L434) | invalid regex | Yes — identical text is raised as `ExpectedToolError` on the next line |
| [`wait.py#L730`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/tools/pane_tools/wait.py#L730) | history-limit trim-risk band | No |
| [`wait.py#L838`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/tools/pane_tools/wait.py#L838) | no match before timeout | Yes — `WaitForTextResult.outcome == "timeout"` plus `elapsed` |

The helper is [`_maybe_log` at `wait.py#L124-L143`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/tools/pane_tools/wait.py#L124-L143); its own docstring names the wire method at [`wait.py#L133`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/tools/pane_tools/wait.py#L133).

## Why stderr rather than a result field

The two tests covering this behavior state the intent: the warnings exist *"so MCP client log panels record the cause independent of the tool result"* ([`test_pane_tools.py#L3997-L3999`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/tests/test_pane_tools.py#L3997-L3999)) and for *"human log readers"* ([`test_pane_tools.py#L4040-L4044`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/tests/test_pane_tools.py#L4040-L4044)). That is an operator/observability role, which is precisely what the spec redirects to stderr — and under stdio, MCP clients surface server stderr in those same log panels.

Promoting them to `WaitForTextResult` fields instead would contradict that model's stated design constraint: *"Field count is load-bearing. `outputSchema` is re-sent on every request of every session, so a field that no agent branches on is a permanent tax."* ([`models.py#L243-L246`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/models.py#L243-L246)). Two of the three warnings are already redundant with the tool result or the raised error, and the third is documented as a caveat in the tool docstring at [`wait.py#L555-L561`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/tools/pane_tools/wait.py#L555-L561), where the agent meets it at tool-selection time.

## Proposed change

- Replace `_maybe_log` with stderr logging through the existing module logger; drop the transport-closed suppression, which has no analogue for stdlib logging.
- Drop the now-unused `ctx` parameter from `_compile_patterns`.
- Reword the `ctx.warning` reference in the `wait_for_text` docstring at [`wait.py#L559`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/tools/pane_tools/wait.py#L559).
- Rewrite the two tests to assert on captured log records rather than a recording context stub.
- Add a `MIGRATION` note: MCP clients that displayed these three warnings in a log panel sourced from `notifications/message` will now receive them on stderr.

`ctx.report_progress` at [`wait.py#L114`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/tools/pane_tools/wait.py#L114) is **not** affected. Progress notifications are not deprecated, and the `2026-07-28` draft explicitly keeps them flowing on the response stream of the request they relate to.

## Out of scope

Checked and unaffected: no Roots, no Sampling, no `includeContext`, no HTTP+SSE transport, and no Dynamic Client Registration anywhere in `src/` — the server is stdio-only at [`server.py#L468`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/server.py#L468).

Two adjacent observations, deliberately not bundled here:

- `_meta` keys minted at [`middleware.py#L313-L343`](https://github.com/tmux-python/libtmux-mcp/blob/v0.1.0a19/src/libtmux_mcp/middleware.py#L313-L343) (`error_type`, `expected`, `suggestion`) are bare names. That is legal — the prefix is optional — but reverse-DNS namespacing is the spec's `SHOULD`, and bare names risk collision with future MCP-reserved keys.
- [SEP-2596](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2596) obliges Tier 1 SDKs to emit a runtime `DeprecationWarning` when a deprecated feature is exercised. `pyproject.toml` currently silences `DeprecationWarning` for `libtmux.*`, `libtmux_mcp.*`, and `tests`, which would mask exactly that signal once the SDK starts emitting it.

tony:
Closes #109.

MCP deprecated the protocol-level Logging capability in [SEP-2577](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577), effective with the `2026-07-28` revision, naming stderr as the migration path for stdio servers. `wait_for_text` was the only place this server exercised it: three `ctx.warning` calls reaching the wire as `notifications/message`. They are now ordinary `libtmux_mcp.*` log records.

Scoped to exactly that. The `logging` **capability** is advertised by FastMCP's own handshake, not by anything in this repo — the three call sites are the only part that is ours to remove.

## Checked against the full registry

All six entries in the [deprecated features registry](https://modelcontextprotocol.io/specification/draft/deprecated), against `src/` and `tests/`:

| Feature | Used here? |
| --- | --- |
| **Logging** | **Yes** — `wait.py:434`, `:730`, `:838` via `_maybe_log` |
| Roots | No |
| Sampling | No |
| `includeContext: "thisServer"/"allServers"` | No — the one `include_context=False` hit is pydantic's `ValidationError.errors()` |
| HTTP+SSE transport | No — stdio-only (`fastmcp.json:9`, `server.py:468`) |
| Dynamic Client Registration | No — no auth surface at all |

## FastMCP will not flag this for you

Worth recording, because it rules out waiting for a warning. In `fastmcp` 4.0.0a2 the two SEP-2577 features are handled differently:

- `ctx.sample()` emits a warn-once `FastMCPDeprecationWarning` (`fastmcp_slim/fastmcp/server/context.py:113`, fired at `:1092` and `:1209`).
- The logging path warns **nothing** at runtime. `send_log_message` carries only a static `# ty: ignore[deprecated]` (`context.py:1659`) with the comment *"Deprecated upstream in SDK v2 but deliberately kept per compat directive; removed with the multi-round-trip follow-up."*

So an SDK upgrade would keep these three calls working silently until MRTR removes them. This was findable only by reading the registry.

## Behavior

`ctx.report_progress` is untouched — progress is not deprecated, and the `2026-07-28` draft explicitly keeps it flowing on the response stream of the request it relates to.

No result field was added. Two of the three warnings are already in the tool result — an invalid regex raises a tool error carrying the identical text, and a timeout is `outcome == "timeout"`. The third, the history-limit trim-risk band, is a diagnostic; `WaitForTextResult` documents field count as load-bearing (`outputSchema` is re-sent every request), so a field no agent branches on would be a permanent tax.

`_compile_patterns` becomes synchronous, since removing the `await` left it with nothing to await.

## Verification

`ruff check`, `ruff format --check`, `mypy` (65 files), `pytest --reruns 0` (801 passed), `just build-docs` clean.

The four migrated tests were checked with a negative control rather than trusted for passing: removing the timeout `logger.warning` from the source turns `test_wait_for_text_warns_on_timeout` red with `messages = []`. Note for future readers — `caplog.at_level(..., logger=...)` does not scope capture (records reach `caplog.records` by propagation to root), so aiming the fixture at a bogus logger name does *not* fail these tests. The `logger=` argument is kept for level-setting and to match the suite's convention.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,25 @@
 _Notes on upcoming releases will be added here_
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### Breaking changes
+
+**{tooliconl}`wait-for-text` warnings move from MCP log notifications to stderr**
+
+MCP deprecated the protocol-level Logging capability in
+[SEP-2577](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577),
+effective with the `2026-07-28` revision, naming stderr as the migration path
+for stdio servers. {tooliconl}`wait-for-text` was the only tool that used it —
+an invalid regex, a timeout, and the history-limit trim-risk band each emitted a
+`notifications/message` warning. All three are now ordinary `libtmux_mcp.*` log
+records on stderr, where every other libtmux-mcp record already goes.
+
+A client that subscribed to MCP log notifications no longer receives these three
+over the protocol. Nothing an agent can act on is lost: the invalid regex is
+still raised as a tool error, and the timeout is still reported as
+{attr}`~libtmux_mcp.models.WaitForTextResult.outcome`. This supersedes the
+0.1.0a19 note that the trim-risk signal is "delivered only as an MCP warning
+notification". (#110)
+
 ### Documentation
 
 #### Caller identity fields are described (#105)
--- a/MIGRATION
+++ b/MIGRATION
@@ -19,6 +19,37 @@ for the full release log.
    [tracker]: https://github.com/tmux-python/libtmux-mcp/discussions
 ```
 
+## libtmux-mcp 0.1.x (unreleased)
+
+### `wait_for_text` warnings arrive on stderr, not as MCP log notifications
+
+MCP deprecated the protocol-level Logging capability in [SEP-2577], effective
+with the `2026-07-28` revision, and names stderr as the migration path for
+stdio servers. `wait_for_text` was the only tool using it.
+
+Three warnings changed channel — an invalid regex, a wait that timed out, and
+polling inside tmux's history-limit trim-risk band. Wording and timing are
+unchanged; only the transport differs.
+
+#### Before
+
+An MCP `notifications/message` at level `warning`.
+
+#### After
+
+An ordinary `libtmux_mcp.tools.pane_tools.wait` log record on stderr. Under the
+stdio transport the client already captures that stream, so most log panels
+show them as before.
+
+If you read these through an MCP client's log handler, switch to the server
+process's stderr. If you drive the tools from Python, `caplog` or any handler
+on the `libtmux_mcp` logger sees them.
+
+This supersedes the 0.1.0a19 note below stating that the trim-risk signal is
+delivered only as an MCP warning notification.
+
+[SEP-2577]: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577
+
 ## libtmux-mcp 0.1.0a19 (2026-07-25)
 
 ### `wait_for_text` takes `patterns`, and `wait_for_content_change` is gone
--- a/docs/tools/pane/wait-for-text.md
+++ b/docs/tools/pane/wait-for-text.md
@@ -64,8 +64,8 @@ rather than rejected, so this can be lower than what you asked for.
 
 Matching is best-effort once polling enters tmux's history-limit trim-risk band,
 because older scrollback can be discarded while the wait is active. The server
-reports that as an MCP warning notification; use {tooliconl}`wait-for-channel`
-for deterministic command completion.
+logs that to stderr; use {tooliconl}`wait-for-channel` for deterministic
+command completion.
 
 ```{fastmcp-tool-input} pane_tools.wait_for_text
 ```
--- a/src/libtmux_mcp/tools/pane_tools/wait.py
+++ b/src/libtmux_mcp/tools/pane_tools/wait.py
@@ -118,31 +118,6 @@ async def _maybe_report_progress(
         return
 
 
-_LogLevel = t.Literal["debug", "info", "warning", "error"]
-
-
-async def _maybe_log(
-    ctx: Context | None,
-    *,
-    level: _LogLevel,
-    message: str,
-) -> None:
-    """Call the matching ``ctx.{level}`` if a Context is available.
-
-    Sibling to :func:`_maybe_report_progress` for client-visible log
-    notifications (``notifications/message`` in MCP). Same suppression
-    contract: silent only when the transport is gone, propagating
-    everything else so programming errors stay loud.
-    """
-    if ctx is None:
-        return
-    method = getattr(ctx, level)
-    try:
-        await method(message)
-    except _TRANSPORT_CLOSED_EXCEPTIONS:
-        return
-
-
 # ---------------------------------------------------------------------------
 # Timeout-bounded tmux reads
 # ---------------------------------------------------------------------------
@@ -412,13 +387,12 @@ async def _first_pane_of_window(
 # ---------------------------------------------------------------------------
 
 
-async def _compile_patterns(
+def _compile_patterns(
     values: list[str],
     *,
     label: str,
     regex: bool,
     match_case: bool,
-    ctx: Context | None,
 ) -> list[re.Pattern[str]]:
     """Compile one pattern list, raising ``ExpectedToolError`` on bad input."""
     flags = 0 if match_case else re.IGNORECASE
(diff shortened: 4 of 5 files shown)


tony:
### Code review

Found 2 issues:

1. The new breaking-changes paragraph has no `(#110)` PR ref (AGENTS.md says "**PR refs `(#NN)`** sit at the end of each deliverable's prose paragraph, not on every sentence.")

Every prior deliverable in the file carries its own ref, added before merge — `(#100)`, `(#98)`, `(#91)`, `(#89)`. This paragraph ends at "delivered only as an MCP warning notification". with none.

https://github.com/tmux-python/libtmux-mcp/blob/5dbc028f2d542f6e9c5e849b3a296812ce3cebc8/CHANGES#L20-L26

2. `WaitForTextResult.outcome` is cited with plain backticks instead of the `{attr}` role (AGENTS.md says "Always link autodoc'd APIs. Any class, function, exception, attribute, or tool slug that has its own rendered page must be cited via the appropriate role (`{class}`, `{func}`, `{exc}`, `{attr}`, `{tooliconl}`) — never with plain backticks.")

Precedent in the same file: `` {attr}`~libtmux_mcp.models.SessionInfo.active_pane_id` ``. The same shape was flagged on #100 for `RunCommandResult.effective_timeout`.

https://github.com/tmux-python/libtmux-mcp/blob/5dbc028f2d542f6e9c5e849b3a296812ce3cebc8/CHANGES#L22-L25

Checked and not flagged: no correctness bugs. Both `_compile_patterns` call sites drop `await`/`ctx=` cleanly, the two lazy `logger.warning` sites have matching placeholder/arg counts, `_TRANSPORT_CLOSED_EXCEPTIONS` is still used by the untouched `_maybe_report_progress`, and all four rewritten tests were traced to their origin commits and still guard the same regressions. The removal of `_maybe_log`'s transport-closed suppression is sound — that guard existed because `ctx.*` does async I/O over the MCP transport, which `logging.Handler.emit()` does not.

Separately, `docs/topics/logging.md` claims FastMCP forwards stdlib log records to clients via the MCP logging capability. That is false against fastmcp 3.4.4 — only explicit `Context.log`-family calls reach the wire — and it predates this PR, so it belongs in its own change rather than here.

PR #114, opened 2026-08-09
Track the FastMCP 4 prerelease line

tony:
## Summary

- **Migrate** the FastMCP requirement to the 4.x prerelease line as a floor (`>=4.0.0b2,<5.0.0`) rather than an exact pin, so installs pick up 4.0.0 when it ships without another release here. FastMCP 4 brings MCP Python SDK v2 with it; no tool, resource, or prompt changes shape, and the server keeps serving whichever protocol era each client asks for.
- **Keep** `tmux://` reads working for tmux identifiers that look like filesystem paths. FastMCP 4 screens templated resource parameters before a handler runs, and its absolute-path check rejects both the socket paths `list_servers` reports and the `<letter>:<rest>` session names tmux accepts — while the equivalent tools resolve them fine.
- **Add** resource coverage for those shapes across every template, plus a guard that a `..` socket name is still refused.
- **Bind** the compatibility reference to the manifest, so the dependency table can no longer drift from `[project].dependencies`.
- **Retire** FastMCP's camelCase compatibility shims in the test suite, with enforcement, so a read that depends on a scheduled-for-removal alias fails the test that made it.
- **Replace** this package's one FastMCP internal import with the documented path, and correct three places that described an error-code transform that never happened.

## Try this branch

Two things differ from the installation methods on the front page. The source is a git ref rather than PyPI, and the FastMCP requirement resolves to a **prerelease**. The prerelease needs no opt-in flag — PEP 440 admits one when only a prerelease satisfies the specifier — but a uv `exclude-newer` cooldown will hide `fastmcp 4.0.0b2` until it ages past the window, which is what the `UV_NO_CONFIG` / `--exclude-newer-package` forms below are for. Drop them once the cooldown lapses, or if you have none configured.

Run it straight from the branch with uvx:

```console
$ uvx --no-config --from git+https://github.com/tmux-python/libtmux-mcp@fastmcp-v4.x libtmux-mcp
```

Keep your cooldown for every other package by exempting only these two distributions. Naming `fastmcp` alone is not enough — the resolver then stops on `fastmcp-slim`:

```console
$ uvx \
    --exclude-newer-package fastmcp=false \
    --exclude-newer-package fastmcp-slim=false \
    --from git+https://github.com/tmux-python/libtmux-mcp@fastmcp-v4.x \
    libtmux-mcp
```

pipx resolves through uv when its uv backend is available, so it inherits the same cooldown. `UV_NO_CONFIG` clears it; `PIPX_DISABLE_UV` does not:

```console
$ UV_NO_CONFIG=1 pipx run --spec git+https://github.com/tmux-python/libtmux-mcp@fastmcp-v4.x libtmux-mcp
```

pip has no cooldown of its own, so it needs nothing extra:

```console
$ pip install git+https://github.com/tmux-python/libtmux-mcp@fastmcp-v4.x
```

From a local checkout, `UV_FROZEN` stops uv re-resolving against the cooldown — the committed lockfile already pins the right versions:

```console
$ UV_FROZEN=1 uv --directory ~/work/python/libtmux-mcp run libtmux-mcp
```

### Registering it with an agent

Substitute one of the commands above for `<CMD>`. Give the server a distinct name if you already have `tmux` registered against a release, so the two cannot collide:

```console
$ claude mcp add tmuxlab -- uvx --no-config --from git+https://github.com/tmux-python/libtmux-mcp@fastmcp-v4.x libtmux-mcp
```

```console
$ codex mcp add tmuxlab -- <CMD>
```

```console
$ grok mcp add --scope user tmuxlab -- <CMD>
```

For the JSON-configured clients — Claude Desktop, Cursor, Antigravity — the cooldown bypass rides as an `env` member:

```json
{
  "mcpServers": {
    "tmuxlab": {
      "command": "uvx",
      "args": ["--from", "git+https://github.com/tmux-python/libtmux-mcp@fastmcp-v4.x", "libtmux-mcp"],
      "env": { "UV_NO_CONFIG": "1" }
    }
  }
}
```

To exercise it without touching a real config, Claude Code takes a throwaway one for the session:

```console
$ claude --mcp-config ./lab.json --strict-mcp-config --permission-mode bypassPermissions
```

Codex reads `CODEX_HOME`, Grok reads `GROK_HOME`, Gemini and Cursor read a project config, and Antigravity takes `--gemini_dir`.

### The check that exercises what this PR changes

Point the server at a scratch socket and create the session name FastMCP 4's screen rejects:

```console
$ tmux -L mcp-target new-session -d -s 'a:1'
```

Then ask the agent to read the **resource**, not the tool — the tool path was never affected:

> Read the MCP resource `tmux://sessions` and list the session names.

On this branch the read returns `a:1`. Against FastMCP 4 without this change it fails with `Resource not found`, while `get_session_info(session_name="a:1")` still resolves the same session — the disagreement between the two views is what this closes.

### Dependency

**`pyproject.toml`**: FastMCP moves to `>=4.0.0b2,<5.0.0`. No `[tool.uv]` block and no `exclude-newer-package` entries are added — FastMCP pins `fastmcp-slim` in its own metadata, so no constraint is needed, and cooldown exemptions are a local-machine concern that would otherwise be committed as policy against this project's most security-sensitive third-party dependencies.

### Resource screening

**`src/libtmux_mcp/server.py`**: Turns off the absolute-path screen for resource parameters, keeping traversal and null-byte screening.

**`tests/test_resources_integration.py`**: Reads every template with an absolute socket path and a path-shaped session name, checks the resource and tool views agree on such a session, and pins that a `..` socket name never reaches tmux.

### Documentation drift

**`tests/docs/test_compatibility.py`**: Compares the documented dependency ranges against the manifest as `SpecifierSet` values, so the doc stays free to space them differently, and fails when a runtime dependency has no row or a row has no dependency.

**`docs/reference/compatibility.md`**: States the floors the manifest actually requires.

### Deprecation debt

**`conftest.py`**: Disables FastMCP's camelCase bridge for the suite, set through the environment before anything imports FastMCP so the subprocess-based tests inherit it.

**`tests/conftest.py`**: Adds `wire_input_schema`, a sibling of the existing `wire_annotations`, reading a tool's argument schema under its protocol name.

## Design decisions

**A prerelease floor, not an exact pin.** `fastmcp==4.0.0b2` installs today — PEP 440 admits a prerelease when only a prerelease satisfies the specifier — but it is sticky: once 4.0.0 ships, every consumer is still on a beta, and the pin becomes a hard resolver conflict with anything asking for `fastmcp>=4.0.0`. A floor self-migrates. Verified on the released line, where `fastmcp>=3.4.0b1,<4` resolves to `3.4.6` rather than the beta.

**`reject_absolute_paths=False`, not `exempt_params` or `resource_security=None`.** The SDK docstring suggests exempting parameters, and both alternatives fix the regression. Neither is right here. `socket_name` arrives through the `{?socket_name}` query, which URI normalization does not touch — unlike path segments, where `..` is collapsed before the handler runs — and it reaches `tmux -L` unvalidated, where a `../..` prefix really does place the socket outside the tmux socket directory:

```console
$ tmux -L '../../tmp/escaped' new-session -d 'sleep 30'
```

An absolute `-L` value cannot escape, because tmux concatenates rather than resolves. So the absolute-path check has no true positive here and two false positives, while the traversal check has a live one. Exempting the parameter, or disabling screening wholesale, would discard the check that still means something. `test_socket_name_traversal_never_reaches_tmux` fails under both alternatives and passes under the chosen one, so the decision is pinned by a test rather than a comment.

**Wire names over Python attribute spellings.** The camelCase migration reads schemas through `model_dump(by_alias=True)` rather than the SDK's `input_schema`. The Python spelling differs across SDK majors while the wire name is what the protocol guarantees and what a client actually sees, so this survives the next rename. It is the idiom `wire_annotations` already established here.

**Enforcement through the environment, not the settings object.** FastMCP builds its settings singleton at import, so the toggle has to precede it; and the schema tests spawn subprocesses through `os.environ.copy()`, which inherits an environment variable but not an attribute assignment. Neither existing gate catches a read that depends on the shims: the type checker sees `Any`, because the SDK objects arrive through `Client.__aenter__`, which is unannotated in both FastMCP 3 and 4; and the deprecation warning fires once per attribute for a whole run.

## Verification

No FastMCP internal module remains imported:

```console
$ rg -n 'fastmcp\.tools\.base|fastmcp\.server\.proxy|fastmcp\.server\.openapi|fastmcp\.server\.apps' src/ tests/
```

No camelCase SDK attribute read survives (matches in prose and the `by_alias` dump key are expected; an attribute access is not):

```console
$ rg -n '\.inputSchema|\.outputSchema|\.isError|\.structuredContent|\.mimeType|\.nextCursor' src/ tests/
```

No surface removed in FastMCP 4 is used:

```console
$ rg -n 'ctx\.sample|list_roots|import_server|as_proxy|exclude_args|sampling_handler|task=True' src/
```

The manifest carries no committed cooldown exemption for the FastMCP or MCP distributions:

```console
$ rg -n -A 30 'exclude-newer-package' pyproject.toml | rg '^\s*(fastmcp|fastmcp-slim|mcp|mcp-types) = '
```

## Test plan

- [x] `uv run ruff check . --fix --show-fixes` — clean
- [x] `uv run ruff format .` — no reformatting
- [x] `uv run mypy src tests` — clean under FastMCP 4
- [x] `uv run py.test --reruns 0` — full suite green
- [x] `just build-docs` — builds with no new warning and no broken cross-reference
- [x] `test_resource_read_accepts_path_like_parameters` — every `tmux://` template reads with an absolute socket path and a colon-bearing session name; fails on every template without the screening change
- [x] `test_resource_and_tool_agree_on_path_like_session_names` — a session readable through `get_session_info` is readable through `tmux://`
- [x] `test_socket_name_traversal_never_reaches_tmux` — a `..` socket name is refused and no server is cached for it; fails under both `exempt_params` and `resource_security=None`, which is what makes it a decision record
- [x] `test_compatibility_dependency_floors_match_the_manifest` — fails when the documented range is edited away from the manifest
- [x] `test_camelcase_bridge_is_disabled_for_the_suite` — asserts the bridge is off, so an import-order change fails loudly rather than quietly restoring the safety net
- [x] Reintroducing a camelCase attribute read raises `AttributeError` and fails its test, rather than passing with a warning
- [x] Real Claude Code, over stdio against this branch, reads `tmux://sessions` and returns a session named `a:1` — the shape that the unmodified FastMCP 4 screen rejects
- [x] `uvx`, `pipx`, and `pip` each install and launch the console script from this branch's git ref

## Companion PR

The documentation toolchain needs a matching change before this branch's rendered docs are correct. `sphinx-autodoc-fastmcp` strips FastMCP's auto-appended prompt-argument schema hint by literal prefix match, and FastMCP 4 reworded it, so the boilerplate now renders in the prompts page. The build stays green and emits no warning, so nothing here catches it.

The fix belongs in `gp-sphinx`, and must match **both** wordings rather than swap one for the other — the same released collector serves repositories on either FastMCP major:

| FastMCP | Emitted prefix |
|---|---|
| 3.x | `Provide as a JSON string matching the following schema:` |
| 4.x | `Provide a value matching the following JSON schema:` |

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -38,6 +38,15 @@ or as a bare name carrying only its type.
 
 ### Development
 
+**The test suite runs without FastMCP's camelCase compatibility shims**
+
+FastMCP still answers MCP SDK v1 attribute spellings through warn-once
+shims it plans to remove. Neither gate catches a test that depends on one:
+the type checker sees `Any`, and the warning fires once per attribute for a
+whole run. The suite now disables the shims, so a read that needs them
+fails the test that made it, and tests read argument schemas under the name
+that goes on the wire instead.
+
 #### CI actions updated to current majors
 
 Workflow actions moved to their current major releases: `actions/checkout` v7,
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,28 @@
 _Notes on upcoming releases will be added here_
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### Dependencies
+
+**Minimum `fastmcp>=4.0.0b2`** (was `>=3.4.2`)
+
+libtmux-mcp now tracks the FastMCP 4 prerelease line. The requirement is a
+floor rather than an exact pin, so installs pick up 4.0.0 once it ships
+without another release here. FastMCP 4 builds on MCP Python SDK v2, which
+brings `mcp>=2.0.0` and pydantic 2.12 with it; no tool, resource, or prompt
+changes shape, and servers serve the same protocol eras their clients ask
+for.
+
+**`tmux://` reads keep accepting tmux names that look like paths**
+
+FastMCP 4 screens templated resource parameters before a handler runs.
+Its absolute-path check does not fit tmux identifiers: it rejects the
+socket paths {tooliconl}`list-servers` reports, and reads any
+`<letter>:<rest>` session name — which tmux accepts — as a Windows
+drive-relative path. Both stay readable, so `tmux://` and the equivalent
+tools continue to agree on which sessions and servers exist. Traversal and
+null-byte screening remain on, where they still guard `socket_name` on its
+way to `tmux -L`.
+
 ### Documentation
 
 #### Caller identity fields are described (#105)
--- a/conftest.py
+++ b/conftest.py
@@ -10,6 +10,7 @@ https://docs.pytest.org/en/stable/deprecations.html
 
 from __future__ import annotations
 
+import os
 import shutil
 import typing as t
 
@@ -23,6 +24,21 @@ from libtmux.window import Window
 
 if t.TYPE_CHECKING:
     import pathlib
+
+# Run the suite with fastmcp's camelCase compatibility bridge switched
+# off. fastmcp 4 still answers SDK v1 spellings like `tool.inputSchema`
+# through warn-once shims that are scheduled for removal, and neither
+# gate catches a read that depends on them: mypy sees `Any` (the SDK
+# objects arrive through `Client.__aenter__`, which is unannotated) and
+# the warning fires once per (class, attribute) for the whole run, so
+# one stray read is easy to miss. With the bridge off a surviving read
+# raises `AttributeError` and fails the test that made it.
+#
+# Set before anything imports fastmcp — its settings singleton is built
+# at import — and via the environment rather than the settings object so
+# the subprocess-based tests inherit it through `os.environ.copy()`.
+# `test_camelcase_bridge_is_disabled_for_the_suite` asserts it took.
+os.environ.setdefault("FASTMCP_MCP_CAMELCASE_COMPAT", "false")
 
 pytest_plugins = ["pytester", "sphinx.testing.fixtures"]
 
--- a/docs/reference/compatibility.md
+++ b/docs/reference/compatibility.md
@@ -24,8 +24,8 @@
 
 | Package | Required version |
 |---------|-----------------|
-| [libtmux](https://libtmux.git-pull.com/) | >= 0.55.0, < 1.0 |
-| [FastMCP](https://github.com/jlowin/fastmcp) | >= 3.1.0, < 4.0.0 |
+| [libtmux](https://libtmux.git-pull.com/) | >= 0.62.0, < 1.0 |
+| [FastMCP](https://github.com/jlowin/fastmcp) | >= 3.4.2, < 4.0.0 |
 
 ## MCP clients
 
--- a/docs/reference/compatibility.md
+++ b/docs/reference/compatibility.md
@@ -25,7 +25,7 @@
 | Package | Required version |
 |---------|-----------------|
 | [libtmux](https://libtmux.git-pull.com/) | >= 0.62.0, < 1.0 |
-| [FastMCP](https://github.com/jlowin/fastmcp) | >= 3.4.2, < 4.0.0 |
+| [FastMCP](https://github.com/jlowin/fastmcp) | >= 4.0.0b2, < 5.0.0 |
 
 ## MCP clients
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ include = [
 
 dependencies = [
   "libtmux>=0.62.0,<1.0",
-  "fastmcp>=3.4.2,<4.0.0",
+  "fastmcp>=4.0.0b2,<5.0.0",
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,9 @@ dev = [
   "syrupy>=5.1.0",
   # scripts/mcp_swap.py (PEP 723 script; dep listed here so tests can import it)
   "tomlkit>=0.13",
+  # Arrives transitively via pytest, but tests/docs/test_compatibility.py
+  # imports it directly to compare version ranges rather than strings.
+  "packaging",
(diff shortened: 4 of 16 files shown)


tony:
Superseded by #131, which tracks the released `fastmcp>=4.0.2` line rather than the `4.0.0b2` prerelease this branch targeted. The prerelease-specific installation guidance above no longer applies, and the checks here are red against several weeks of `main`.

Everything on this branch is in #131, rebased onto current `main`: the resource path-screening fix and its reasoning, the compatibility-table binding, the camelCase shim retirement, the public-import move, and the not-found error-code correction. #131 adds `tmux://` argument completion, a wait-tool fix for a context carrying no session, and the upstream write-up for the error code.

Two things worth carrying forward rather than losing with this description. The `reject_absolute_paths=False` decision — chosen over `exempt_params` and `resource_security=None` because `socket_name` still needs traversal screening on its way to `tmux -L` — remains pinned by `test_socket_name_traversal_never_reaches_tmux`. And the `gp-sphinx` companion fix is still outstanding: FastMCP 4's reworded prompt-argument hint now renders seven times on the built prompts page, and the docs build passes anyway.

One decision is reversed rather than carried. This pull request deliberately declined to commit `exclude-newer-package` entries for the FastMCP distributions, on the grounds that a cooldown exemption is a local-machine concern. #131 commits them, and its description says so explicitly alongside 8399624 — the ruff exemption revert that argues the same way this description did. A reviewer who wants them dropped has that precedent and needs only the lock commit reverted.

PR #129, opened 2026-08-30
Make tmux MCP operations bounded and truthful

tony:
## Summary

- **Bounded execution:** Centralize tmux calls behind finite admission, subprocess, output, batch, and cleanup limits without blocking the event loop.
- **Truthful outcomes:** Return exact success, failure, location, identity, and destructive-effect information instead of silent or inferred results.
- **Safe targeting:** Require explicit stable targets where ambiguity is dangerous and protect caller, nested-tmux, and last-pane/session boundaries.
- **Literal input:** Preserve names, paths, commands, filters, hooks, options, and environment values without flag injection or unintended expansion.
- **Reliable observation:** Preserve `capture_since` continuity across reflow and floods while making `wait_for_text` handle repeated lines, stops, pane death, and progress accurately.
- **Broader tool surface:** Add client, environment, hook, buffer, break, join, move, and server operations with bounded resources and redacted audit output.

## Review note

This proposal includes every commit currently reviewed in #125 plus the later `improvements-00` work. A new main-targeted PR overlaps that open PR.

## Compatibility

- Invalid, ambiguous, or inferred targets previously accepted by some tools now fail with actionable errors.
- Result schemas and effect metadata become stricter and more explicit.
- Configuration swapping preserves JSONC comments and can target local checkouts or pull requests across supported clients.

## Test plan

- [x] Net patch passes `git diff --check`.
- [ ] Rebase onto the substantially advanced `main` branch.
- [ ] Resolve the Python 3.14/tmux 3.4 `break_pane` failure before rerunning the matrix.
- [ ] Rerun pytest, Ruff, mypy, documentation, and real-tmux isolation gates.


tony:
The branch-local research notes are archived here before removing `notes/*` from the PR commit history. Machine-specific values and internal process attribution are generalized; exact originals remain in the requested local backup.

<details>
<summary><code>notes/spikes/control-mode-for-waiting.md</code></summary>

# Spike: control mode instead of polling for `wait_for_text`

**Verdict: do not adopt it for waiting.** It is the right primitive and the wrong trade for this server. Three kill-questions were asked before any code; two came back with costs that outweigh what polling costs today.

Not a code change. Written so the question does not have to be reopened from scratch.

## Why it was asked

`wait_for_text` polls, because tmux has no hook that fires on pane output — the `notify_*` set in `notify.c` covers session, window and client lifecycle only. Polling costs two tmux client spawns per tick: about 14% of one core at the default 0.05 interval, 3.3% at 0.25, hard capped at 120 s per call.

Control mode (`tmux -C` / `-CC`) is the real push primitive. `%output` arrives when a pane writes, with no hook, no `pipe-pane` slot to contend for, and no shell-expansion surface.

## Q1 — does attaching a control client perturb what this server reports?

**Yes, in two of the three places checked.** Measured against tmux 3.7c, `-C` and `-CC` behaving identically:

| observed | before | attached | after exit |
| --- | --- | --- | --- |
| `session_attached` | 0 | **1** | 0 |
| `list-clients` | empty | **one row, `client_control_mode=1`** | empty |
| window size | 80x24 | 80x24 | 80x24 |

Sizing is safe by construction: `ignore_client_size` (`resize.c:91`) returns true for a control client that has set neither `CLIENT_SIZECHANGED` nor `CLIENT_WINDOWSIZECHANGED`, so it is excluded from `recalculate_size`.

The other two are not. `CLIENT_CONTROL` is absent from `CLIENT_UNATTACHEDFLAGS` (`tmux.h:2137`), so `resize.c:448` counts a control client as attached like any other.

This is the finding that decides the spike. `session_attached` and `list_clients` are not internals here — they are answers this server gives agents, and "is anyone using this session?" would become wrong for the duration of every wait. `list_clients` could filter our own client out; `session_attached` is tmux's own count and cannot be un-incremented without lying about it.

The first attempt at this measurement reported no perturbation at all, because the control client had died instantly on `tcgetattr failed: Inappropriate ioctl for device` — a null result identical in shape to the healthy one. It needs a real pty.

## Q2 — lifecycle

A control client is a long-lived child holding a pty, which is a heavier commitment than the killable one-shot subprocess the wait path owns now, and it must die with the server: the reap discipline in `_tmux_proc` would have to cover it. tmux can also end it from its own side — see Q3 — so the connection is not a resource the server fully controls.

## Q3 — can output stall or be lost?

**It can be lost, and the default is to lose all of it.** When a pane's oldest unsent block exceeds an age bound (`control_check_age`, `control.c:438`):

- **default**: past `CONTROL_MAXIMUM_AGE` — 300000 in `get_timer()` milliseconds, so **300 s** — tmux sets `exit_message` to `"too far behind"`, flags `CLIENT_EXIT`, and calls `control_discard(c)`. The client is killed.
- **with `pause-after`**: past the configured age the pane is marked `CONTROL_PANE_PAUSED`, `control_discard_pane` frees every buffered block, and `%pause %<id>` is emitted; `%continue` follows on resume.

So the consumer is told it fell behind, and the output it missed is gone either way. For a wait tool that means a missed match reported as a timeout — the false negative that would make this strictly worse than polling, which cannot miss what is still on the grid.

`CONTROL_BUFFER_HIGH` is 8192 bytes and `CONTROL_BUFFER_LOW` 512 (`control.c:131`).

Both mechanisms landed in tmux 3.2 (`CHANGES FROM 3.1c TO 3.2`), at or below this project's 3.2a floor, so no version branching would be needed — this is not why the answer is no.

## What would change the verdict

- A use case that is not `wait_for_text`: streaming a pane to a client deliberately, where an attached client is honest rather than a side effect, and where `%pause` is reportable to the caller instead of being a missed match.
- `session_attached` gaining a way to exclude control clients upstream.

Neither is true today, and polling's measured cost is bounded and has a caller-facing knob. The current design is the right one.

</details>

<details>
<summary><code>notes/upstream-libtmux/README.md</code></summary>

# Upstream libtmux notes

Seven things this MCP server ran into that belong in libtmux rather than here. Written for the maintainer's judgement and **not filed as issues** — opening those is a separate, outward-facing decision.

Each note states what was measured and how, so the claim can be checked rather than believed:

| Note | What it is |
| --- | --- |
| `01-client-fields` | `Session` advertises `client_*` / `buffer_*` format fields it never populates |
| `02-hook-include-inherited` | `show_hook` has no `include_inherited`, so an inherited hook is invisible |
| `03-numeric-operators` | `QueryList` filters have no numeric comparisons, so range queries are done in Python |
| `04-parent-properties-refetch` | `Pane.window` / `Pane.session` re-query tmux instead of using the ids the object already holds — `pane.session` costs two subprocesses, and `Pane.split()` costs eight |
| `05-unresolvable-socket-path` | `Server.socket_path` is `None` unless the caller passed it — by name or by default — so `str(...)` yields `'None'` and every connect fails as `server exited unexpectedly` |
| `06-no-timeout-on-tmux-cmd` | No caller can bound a tmux call; a stalled server holds the worker forever, and 40 accumulated calls wedge the host process |
| `07-test-fixture-orphans-servers` | The pytest fixture unlinks a test server's socket even when the kill did not happen, leaving a live server nothing can reach |

Archived in this PR comment so the findings survive branch-history cleanup without becoming repository files.

</details>

<details>
<summary><code>notes/upstream-libtmux/libtmux-01-client-fields.md</code></summary>

Title: Session exposes client*\* and buffer*\* format fields it never populates

`Session` advertises the whole `client_*` and `buffer_*` format family as attributes, and they are `None` regardless of whether a client is attached.

Measured on libtmux 0.58 / tmux 3.7c, with a real pts client confirmed by `list-clients`:

    == DETACHED ==
      clients attached: (none)
      Session.client_pid  = None
      attributes that are None: 64

    == ATTACHED (real pts client) ==
      clients attached: /dev/pts/N pid=<pid>
      Session.client_pid  = None
      Session.client_name = None
      attributes that are None: 62

So 62 of 230 attributes reachable through `dir(Session)` can only ever answer empty. That matters beyond cosmetics because `QueryList.filter()` resolves keys by `getattr`, so filtering on any of them silently returns nothing -- indistinguishable from "no rows matched".

Separately, three attributes RAISE rather than return `None`:

    attached_pane, attached_window, children  -> DeprecatedError

They are still visible to `dir()`, so introspection hands them out as valid field names, and `keygetter`'s `except Exception: return None` turns the raise into an empty result.

Suggestion: either populate the client fields from `list-clients`, or drop them from the attribute surface so that introspection does not advertise fields that cannot answer.

</details>

<details>
<summary><code>notes/upstream-libtmux/libtmux-02-hook-include-inherited.md</code></summary>

Title: show_hook has no include_inherited (-A) equivalent, and drops tmux's inherited marker

`Server.show_option` accepts `include_inherited`, which maps to tmux's `-A`. `Server.show_hook` does not, and there is no way to ask whether a hook was set at the scope queried or inherited from a wider one.

tmux answers this itself. With a window hook set on `@0`:

    show-options    -w -t @0  pane-focus-in -> pane-focus-in[0]  ...   set here
    show-options -A -p -t %0  pane-focus-in -> pane-focus-in[0]* ...   INHERITED

The trailing `*` is the marker, and it is parsed away before a caller sees it.

Consequence downstream: a hook-listing tool can enumerate what is set at a scope, and a name-targeted lookup resolves with inheritance, but neither can tell the caller which kind of answer it just gave.

Suggestion: add `include_inherited` to `show_hook`/`show_hooks` for symmetry with the option API, and surface the inherited marker rather than stripping it.

</details>

<details>
<summary><code>notes/upstream-libtmux/libtmux-03-numeric-operators.md</code></summary>

Title: QueryList has no ordering operators, and adding them needs numeric coercion

`LOOKUP_NAME_MAP` is contains, endswith, eq, exact, icontains, iendswith, iexact, in, iregex, istartswith, nin, regex, startswith. There is no `gt`, `gte`, `lt` or `lte`, so "sessions with more than one window" is not expressible.

Worth flagging before anyone adds them: every tmux format field arrives as a STRING. `session_windows` is `'2'`, not `2`. A naive ordering operator would compare lexicographically, so `'9' > '10'` and the results would be wrong in a way that looks correct for single-digit values -- which is most test fixtures.

Suggestion: if ordering operators are added, coerce both sides to a number when both parse as one, and document the fallback for values that do not.

Related, and the reason this surfaced: three lookups already fail silently on non-string data. `lookup_in`, `lookup_nin` and `lookup_regex` all guard on `isinstance(data, (str, list))` or `Mapping` and fall off the end to `return False` for a bool. Measured against a three-row set, `__in` and `__nin` both return 0 rows -- every row is either in or not in, so one of those answers is false whichever way you read it -- and `__regex=".*"` matches nothing.

</details>

<details>
<summary><code>notes/upstream-libtmux/libtmux-04-parent-properties-refetch.md</code></summary>

# libtmux: `Pane.window` and `Pane.session` re-fetch what the object already holds

## What happens

`Pane` carries `window_id`, `session_id` and `session_name` as plain fields, populated by the same `-F` query that built the pane. The parent _properties_ ignore them and re-query tmux:

```python
@property
def window(self) -> Window:
    """Parent window of pane."""
    assert isinstance(self.window_id, str)
    from libtmux.window import Window

    return Window.from_window_id(server=self.server, window_id=self.window_id)


@property
def session(self) -> Session:
    """Parent session of pane."""
    return self.window.session
```

Measured, counting `subprocess.Popen` calls on a fresh server, one access each:

```
pane.window_id  (field)      0 tmux subprocesses
pane.window     (property)   1
pane.session    (property)   2      <- goes through .window first
```

`Pane.split()` costs 8 subprocesses for an operation raw tmux performs in one `split-window` (3.4 ms), because the implementation reaches for `self.window` four times and `self.session` twice.

## Why it matters

Downstream, `libtmux-mcp`'s `split_window` measured 13 tmux round trips per call — 10 of them from inside libtmux, 3 from the MCP's own resolve and serialize helpers. At a p90 of 5.5 ms per round trip that is ~70 ms of pure process spawning for one split, and it scales with every tool that touches a parent.

The MCP's own serializer is deliberately written to read the flat fields (`pane.window_id`, `pane.session_id`, `pane.session_name`) rather than the properties, precisely to avoid this — which is evidence the properties are a trap worth removing rather than documenting.

## What memoisation would actually buy

Measured rather than estimated, by caching the two properties per instance at runtime and counting `subprocess.Popen`:

```
Pane.split() as shipped              8 tmux subprocesses
  with parents memoised per instance 4
                                     ^ half, and raw tmux does it in 1

libtmux-mcp split_window as shipped 13 tmux subprocesses
  with parents memoised              9   (30% fewer)
```

Nothing else changed — only `Pane.window` and `Pane.session` cached on the instance. The remaining four inside `split()` are distinct queries rather than repeats.

## Suggested shape

Memoize per object. Building the parent from the fields already present is not available -- a `Pane` holds `window_id` but not the window's other fields, so a full `Window` cannot be constructed without asking. The correctness argument for re-fetching is staleness, but it is not consistently applied: the pane's OWN fields are equally stale and are returned without a refresh, so the property pays for a guarantee its neighbours do not offer.

If staleness is the concern, `refresh()` already exists and is explicit. A property that silently costs a subprocess is the surprising half.

## Not filed

Written for the maintainer's judgement, not opened as an issue.

Spawn counts were reproduced independently during the latency audit.

</details>

<details>
<summary><code>notes/upstream-libtmux/libtmux-05-unresolvable-socket-path.md</code></summary>

# libtmux: a Server cannot report the socket path it will use

## What happens

`Server.socket_path` stores what the caller passed, and nothing else ever sets it. It is `None` on a server built by name and `None` on the default server, even though tmux resolves both to a deterministic path:

```
Server(socket_name='demo-note')    socket_path=None  str()='None'
Server()  # default server         socket_path=None  str()='None'
```

So this is not a `socket_name` quirk. The attribute is `None` unless the caller supplied the very thing they are asking for.

There is no public helper that closes the gap. `socket_path_from_env` exists but lives in `libtmux._internal.env`, and it answers a different question — parsing `$TMUX` for a server you are already inside, not resolving a name for one you are not:

```
public resolver: ImportError cannot import name 'socket_path_from_env' from 'libtmux'
```

## Why it matters

The attribute reads like a question about the server and answers a question about the constructor, so the failure is silent and typed-correct. `str(None)` is `'None'`, a perfectly good four-character string, and every consumer of it connects to a path that does not exist.

Downstream that is not hypothetical. `libtmux-mcp` carries a private `_effective_socket_path` helper for exactly this, with a three-step resolution order — the stored attribute, then `display-message -p '#{socket_path}'` because tmux is authoritative under macOS launchd where `$TMUX_TMPDIR` diverges, then reconstruction from `${TMUX_TMPDIR:-/tmp}/tmux-<euid>/<name>`. Every serializer in the package spells `str(server.socket_path) if server.socket_path else None` rather than `str(...)`.

The sharper cost is in test instruments, where nothing guards. `str(None)` makes every `connect()` fail with `ENOENT`, and a tmux client whose socket is closed under it reports `server exited unexpectedly` — indistinguishable from a genuinely dead server. A proxy fixture written this way forwards nothing while still producing plausible output; the reading it gave at one depth was byte-identical to the reading from a working proxy, so the defect could not be seen in the result it produced. Three independent test instruments reproduced the same failure.

## Suggested shape

A read-only resolved path, not a change to `socket_path` itself — that attribute participates in `__eq__` and `__repr__` (`server.py:2676`, `2683`), so computing it would alter equality between a name-built and a path-built server pointing at the same socket.

Something like a `resolved_socket_path` property, or promoting the reconstruction to a public module function, would let callers ask the question without each one re-deriving `${TMUX_TMPDIR:-/tmp}/tmux-<euid>/<name>`. tmux's own `#{socket_path}` is the authority when the server is live; the reconstruction is what is needed when it is not.

## Not filed

Written for the maintainer's judgement, not opened as an issue.

Reproduced against the pinned libtmux version on this branch; the two blocks above are that run's output, not a transcription.

</details>

<details>
<summary><code>notes/upstream-libtmux/libtmux-06-no-timeout-on-tmux-cmd.md</code></summary>

# libtmux: no caller can bound a tmux call

## What happens

`tmux_cmd.__init__` runs the command and waits with no deadline (`libtmux/common.py`):

```python
self.process = subprocess.Popen(cmd, stdout=..., stderr=..., text=True)
stdout, stderr = self.process.communicate()
```

`communicate()` without `timeout=` returns when tmux answers, and only then. There is no parameter anywhere in the chain -- `tmux_cmd`, `Server.cmd`, `fetch_objs` -- that lets a caller say how long it is prepared to wait.

That is fine against a tmux that answers. A tmux server spinning inside its own event loop accepts the connection and never replies, and then the call never returns.

## Why it matters

A socket with a listener is not a server that answers, and the difference is invisible until something blocks on it. Measured downstream, against a relay that forwards one connection to a real server and stalls every later one:

```
break_pane, healthy server     0.13s, 11 tmux round trips
break_pane, stalls at trip 2   still running at 150s
```

Three further properties make it worse than one slow call:

- **The worker is never returned.** Cancelling the coroutine that awaits the call does not interrupt `communicate()`. Measured: the tmux client is still alive 0.5s, 3s and 8s after `CancelledError`.
- **It accumulates.** Because nothing is returned, hung calls only ever add up. Forty of them -- concurrent, or sequential with a cancel between each -- exhaust anyio's default thread limiter (`total_tokens` is 40 regardless of CPU count) and the host process stops serving _everything_, including requests aimed at completely healthy sockets.
- **No attacker is required.** Forty is reached by an agent behaving correctly: call a tool, give up when it does not return, retry.

## `Server.cmd` is not the only funnel

Worth stating separately, because it defeats the obvious workaround. A downstream consumer cannot fix this with a `Server` subclass: three call sites construct `tmux_cmd` and only one is reachable that way.

| site | reached by a `Server` subclass? |
| --- | --- |
| `server.py:408` `Server.cmd` | yes |
| `neo.py:755` `fetch_objs` | **no** -- module-level, builds `tmux_cmd` directly |
| `common.py:425` `get_version` | **no** |

`fetch_objs` is the QueryList engine behind `Server.sessions`, `Session.windows` and `Window.panes`, so it is the busiest of the three. A subclass bounding `Server.cmd` was measured leaving `break_pane` unbounded for exactly this reason.

## Suggested shape

A `timeout` parameter on `tmux_cmd`, threaded through `Server.cmd` and `fetch_objs`, with a server-level default so a caller can set the policy once. `subprocess.run(..., timeout=)` is enough machinery: it kills and reaps the child before raising, which is the property that stops stalled calls from accumulating -- measured downstream as 1 tmux client mid-call and 0 after the bound fires.

The exception type matters. `Server._fetch_or_empty` catches `LibTmuxException` and returns `[]` for a not-yet-started daemon, so a timeout raised as a `LibTmuxException` would make a _wedged_ server report zero sessions. It needs to be a distinct type.

## Not filed

Written for the maintainer's judgement, not opened as an issue.

Numbers are from a downstream reproduction, not transcribed. The workaround shipped there rebinds `tmux_cmd` in all three modules, with a test that AST-walks the installed libtmux and fails if a fourth call site appears.

</details>

<details>
<summary><code>notes/upstream-libtmux/libtmux-07-test-fixture-orphans-servers.md</code></summary>

# libtmux: the pytest fixture converts a failed kill into an unreachable server

## What happens

`pytest_plugin.py::_reap_test_server` kills the test server and then removes its socket file. The kill is conditional and suppressed; the unlink is neither:

```python
with contextlib.suppress(exc.LibTmuxException, OSError):
    srv = Server(socket_name=socket_name)
    if srv.is_alive():
        srv.kill()  # skipped, or raises and is swallowed

tmux_tmpdir = pathlib.Path(os.environ.get("TMUX_TMPDIR", "/tmp"))
socket_path = tmux_tmpdir / f"tmux-{os.geteuid()}" / socket_name
with contextlib.suppress(OSError):
    socket_path.unlink(missing_ok=True)  # runs unconditionally
```

The unlink recomputes the path by hand, with a comment explaining that `Server(socket_name=...)` never populates `socket_path` — which is the gap in `05-unresolvable-socket-path`, showing up inside libtmux's own code.

So when the kill does not happen, the socket is removed anyway. The result is a live tmux server with no socket file: unreachable by `tmux -L <name> kill-server`, reparented to init, and never reaped.

The gate is `Server.is_alive` (`server.py:287`), which reports a failure to _ask_ as an answer about the server:

```python
def is_alive(self) -> bool:
    try:
        res = self.cmd("list-sessions")
    except Exception:
        return False
    return res.returncode == 0
```

Under load, `self.cmd` can fail because `fork` returned `EAGAIN` or the subprocess timed out. Neither says the server is dead.

## Reproduced

Deterministic, against stock libtmux 0.62.0 from the index:

```
CONTROL: no failure                    is_alive=True   procs 1->0  socket gone  no orphan
Server.cmd raises OSError/EAGAIN       is_alive=False  procs 1->1  socket gone  ORPHAN
Server.cmd raises TimeoutExpired       is_alive=False  procs 1->1  socket gone  ORPHAN
```

The only difference between the control and the failing rows is that the query raises — exactly what the `suppress` above is written to absorb.

A competing explanation was tested and rejected: `list-sessions` against a running server with no sessions was expected to exit non-zero and make `is_alive` return False naturally. It does not — `rc=0`, empty stderr, `is_alive()` True, and the reap is clean. The trigger is the exception path, not the returncode path.

## Observed in the wild

After repeated suite runs:

```
orphaned tmux servers          >1,500
libtmux_test socket files         <100
oldest orphaned server            >13h
parent                   init -- nothing will reap them
RSS                      >5 GB
```

The asymmetry between those first two numbers is the signature: sockets unlinked, servers not killed.

It also accelerates. Every orphan is a live process, more processes make `fork` failures likelier, and more `fork` failures make `is_alive` wrong more often. The age distribution matches a rate that grows rather than a single event.

## Why it matters beyond tidiness

Every timing measurement taken on such a machine is confounded — a suite that has been running for a day carries a load its own fixture created. On a CI runner the same behaviour would accumulate until it hit a process or memory limit, and the failure would surface as an unrelated flaky test.

## Suggested shape

Two changes, either of which breaks the chain:

- Unlink the socket only when the kill is known to have succeeded, or the server is known to be gone. A cleanup that cannot kill should leave the handle in place rather than remove the only way to reach it.
- Give `is_alive` a way to say "could not determine", distinct from "dead". The bare `except Exception: return False` conflates them, and every caller is written as though they were distinguishable.

The docstring's reasoning for suppressing is good and should survive: a finalizer that raises replaces the real test failure with a cleanup error. That argument holds for the unlink. It does not hold for the kill, because a kill that silently fails is not an unactionable cleanup failure — it is a leaked process, made invisible by the next statement.

## Not filed

Written for the maintainer's judgement, not opened as an issue.

The mechanism was reproduced with a control and two triggers; one competing hypothesis was tested and discarded.

</details>

